PR #15, opened 2026-06-05
Add DeepSeek provider integration

matheusht:
## Summary
- add DeepSeek as an accepted OpenAI-compatible LLM provider
- allow HTML snapshot scenarios to run with PERFECTMAN_LLM=deepseek
- fix replay/scheduler cursor behavior that made DeepSeek snapshots look one-sided or stale
- add a DeepSeek example simulation config and update the generated HTML snapshot

## Validation
- pnpm --filter @perfectman/server typecheck
- pnpm --filter @perfectman/server test
- pnpm build
- PERFECTMAN_LLM=deepseek PULSE_COUNT=15 pnpm --dir packages/server exec vitest run src/__e2e__/html-snapshot.e2e.test.ts

DeepSeek snapshot verified: mock=false, deepseek-chat/deepseek-v4 present, no llm_failure, no fallback.


caiotheodoro:
## Audit review

Checked this PR against `main` and against the other open PR (#16, test taxonomy/hygiene). Summary: the idea is right, but this PR is **stale and conflicts — needs a rebase + rework before it can merge**.

### Blocker 1 — stale base
Base is `feat/refinement`, which was merged long ago (PR #14). The branch is 8 behind `main`, and `main` has since **rewritten `pulse-scheduler.ts`** (`12cb1f4`: simulated clock + rolling context window). `git merge-tree` confirms a conflict with `main` on `docs/simulation-snapshot.html`, and with #16 on `html-snapshot.e2e.test.ts` + `pulse-scheduler.ts`.

### Blocker 2 — scheduler cursor fix targets the OLD scheduler
The `getAfter(sim.id)` no-cursor change + `agentStates.set(...)` propagation in `pulse-scheduler.ts` were written against the pre-rewrite scheduler. Main's scheduler now uses `pulseStartEventId` cursors + a `CONTEXT_WINDOW_PULSES_LIMIT` rolling window, so this needs to be re-derived against current code, not merged blindly. Also need to verify whether the anti-drift behavior it introduced is already covered by the rewrite.

### Blocker 3 — shared-file conflicts with #16
- `docs/simulation-snapshot.html`: #16 **deletes** it (evidence moves to gitignored `tmp/`). Recommend resolving as delete — the regenerated HTML here is superseded.
- `html-snapshot.e2e.test.ts`: #16 rewrites the e2e family into one parameterized suite. The new assertions here (thinking ≤ agentsCalled, multi-agent-after-opening guard) should be **ported** into that rewritten suite rather than re-applied as a conflict resolution.

### Not a blocker — overlap with eval layer
DeepSeek **already works at the eval layer on `main`** (evidence committed under `docs/eval/evidence/deepseek`, routed via the eval package's own `PERFECTMAN_LLM_PROVIDER=deepseek`). The remaining unique value here is the **server-runtime** `providerType: "deepseek"` in `llm-config.ts`/`simulation-config.ts` + the example config — worth keeping after rebase.

### Ask
- Should I rebase this onto `main` + port the scheduler fix against the new scheduler, and the e2e assertions into #16's suite? The follow-up eval-premise PR (#16 + gaps) will land on top regardless.

PR #16, opened 2026-08-19
test: testing strategy, hygiene gate, and test-quality refactor

yoarajota:
## Summary

Introduces a testing taxonomy + hygiene gate and a codebase-wide test-quality pass for `perfectman`, per `docs/testing-strategy.md`. The suite was restructured from self-duplicating ~65 files to a layered, assertion-verified set, a drift-prone test double was removed, and an enforceable `pnpm test:gate` now protects it. `docs/testing-strategy.md` is a stale-proof agent guideline (taxonomy, Q1–Q11 standards, enforcement, review checklist).

## What's in this PR

**Strategy & tooling**
- `docs/testing-strategy.md` — honeycomb-style taxonomy (`contract`/`integration`/`e2e` + separate eval benchmark), 11 quality standards, resolved structural decisions, and an agent-facing review checklist. Written as a durable guideline (no counts/paths that rot).
- `scripts/audit-tests.mjs` — hygiene scanner promoted to a CI gate (`pnpm test:gate` / `test:all`). Fails on `.only`/`.skip`, `as any`, `@ts-ignore`, `eslint-disable`, dead id-truthy, weak-terminal `toBeDefined`, and over-cap files.

**Quality fixes (each verified against production code)**
- Removed dead assertions (`expect(id).toBeTruthy()` on generated IDs), tautologies, and redundant shape guards (fields fully covered by TS types).
- Collapsed 4 near-identical e2e suites into one parameterized suite (~38 → 23 tests, no behavior lost).
- Split one 47-`it` SQLite test file into 6 per-repository files + shared `sqlite-test-helpers.ts`.
- Constants: per-table validity checks condensed 31 → 13 without losing coverage.
- Corrected a factually wrong test claiming `channel_created` payload carried `privateMotiveSummary` (it carries `channelType`/`invitedAgentIds`).

**Removed a drift-prone test double**
- Added a `stepResolver` seam to `PulseScheduler` (default `runEngineStep`; production unchanged) so commit-ordering is testable against the **real** scheduler.
- Deleted the hand-rolled `PulseOrchestrator` double whose event payloads had drifted from production; its blocked/delayed branching was already covered by real `intent-resolver` tests. Ordering (`memory_written`, no-op-before-LLM) now covered through the real scheduler.

[...]

PR #17, opened 2026-08-19
feat(eval): heuristic LLM-as-judge at high temp with per-turn narrative cohesion

caiotheodoro:
## What this adds

Closes the eval-premise gaps from the testing-strategy discussion (heuristic
LLM-as-judge, temperature up, per-turn eval, narrative cohesion). Stacks on
top of #16 (test taxonomy/hygiene).

### Judge temperature via env — default HIGH
`packages/eval/src/cli/bench.ts` now reads `PERFECTMAN_JUDGE_TEMPERATURE`;
the judge **defaults to 1.0** instead of the hardcoded `0`. The premise is a
*heuristic* judge: varied, creative reads expose cohesion/voice failures that a
strict low-temp judge flattens. Calibration runs set it to 0 explicitly
(documented in `docs/eval/README.md` and the code).

### `narrative_cohesion` rubric axis
`packages/shared/src/scenarios/rubrics.ts` — new `ROLEPLAY_AXES` entry with
anchors (1=contradicts itself … 5=arcs with callbacks/escalation/shifted
meaning) plus a `roleplay-v1` target of 3.5.

### Per-turn eval path
`packages/eval/src/judge/judge.ts`:
- `llmJudgePerTurn(scenario, events, config, maxTurnSamples=8)` — one
  whole-transcript call for all axes, then consecutive content-bearing turns
  (grouped by pulse) are each scored on cohesion against the preceding turn;
  the mean overrides `narrative_cohesion`.
- `ruleJudge` maps a proxy for the new axis so the offline baseline doesn't gap it.
- `--per-turn` flag on the bench CLI.

### Tests
`packages/eval/src/test/judge-signals.test.ts` — rule-judge axis bounds,
per-turn aggregation (mock fetch), and graceful fallback when per-turn calls
fail (whole-transcript default is kept).

## Validation
- `pnpm -r typecheck` clean
- `pnpm --filter @perfectman/eval test` (24) · `shared` (59) · `engine` (289) · `server` non-e2e (356) green
- `pnpm test:gate` hygiene gate passes

## Relationship to open PRs
- Base: `refactor/test-taxonomy-and-hygiene` (#16)
- Commented on #15 (DeepSeek server-runtime): stale base + conflicts; DeepSeek
  already works at the eval layer on main — that PR's remaining value is the
  server-runtime provider type + example config.

### DIFF
--- a/docs/eval/README.md
+++ b/docs/eval/README.md
@@ -53,9 +53,11 @@ pnpm --filter @perfectman/eval bench --scenarios v1_mention_reply,motive_gossip 
 The report (`bench-report-v1`) contains per-scenario signal/probe/judge
 results, probe averages, judge axis means vs targets, category splits, and
 the judge calibration report. Failing reports are committed, never hidden.
-`--per-turn` adds a `narrative_cohesion` axis: consecutive content-bearing
-turns are each scored against the turn before them, and the mean overrides
-the whole-transcript default (see `packages/eval/src/judge/judge.ts`).
+The `narrative_cohesion` axis lives on the `roleplay-v1` rubric (anchored for
+turn-to-turn thread, callbacks, escalation, shifted meaning). By default the
+LLM judge scores it from the whole transcript; `--per-turn` (LLM judge only)
+replaces that with the mean of per-turn scores — each sampled content-bearing
+turn is scored against the turn before it (see `packages/eval/src/judge/judge.ts`).
 
 ## Current baseline (mock, 123 tasks)
 
--- a/docs/eval/README.md
+++ b/docs/eval/README.md
@@ -37,6 +37,14 @@ PERFECTMAN_LLM_BASE_URL=http://localhost:11434/v1 \
 PERFECTMAN_LLM_MODEL=qwen3:8b \
 pnpm --filter @perfectman/eval bench --mode local --judge llm --out out/bench-local.json
 
+# heuristic LLM-as-judge with per-turn narrative-cohesion scoring.
+# The judge defaults to HIGH temperature (PERFECTMAN_JUDGE_TEMPERATURE) —
+# varied, creative reads expose cohesion/voice failures a strict low-temp
+# judge misses. Set it to 0 for deterministic calibration runs.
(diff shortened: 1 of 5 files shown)


caiotheodoro:
## Independent review + fixes (separate-context cursor agent)

Dispatched a fresh reviewer (separate context, tool `cursor`) against `a5e9fad` vs the stack base. Verdict was **needs changes** — 4 real issues, now fixed in `2343d92`:

1. **Sampling off-by-one** — loop started at `i=0`, so `step>1` skipped the opening pair and under-sampled long transcripts (e.g. 10 pulses → 4 pairs, never `(0,1)`). Now starts at `i=1`: strided adjacent pairs include `(0,1)`. Added a 10-pulse strided-sampling test asserting which pairs are fetched.
2. **Axis stamped onto wrong rubrics** — `llmJudgePerTurn` wrote `narrative_cohesion` on every scene including edge-chaos/behavioral. Now guarded: if the rubric lacks the axis, per-turn calls are skipped entirely (test added, asserts 1 call only).
3. **JSON shape mismatch in `scoreCohesion`** — accepted only `{narrative_cohesion:n}`; a wrapped `{axes:{...}}` parsed as 3 and overwrote a good whole-transcript score. Now accepts both shapes.
4. **Aggregation test didn't test aggregation** — mock returned the same score per-turn and the whole-transcript call was `{narrative_cohesion:...}` (ignored by `llmJudge`); assertion `5` couldn't distinguish mean from coincidence. Rewrote with a real `{axes:{...}}` whole-transcript payload and distinct per-turn scores `3,5 → 4`.

Also: empty `PERFECTMAN_JUDGE_TEMPERATURE` treated as unset (`Number('')=0` footgun), and README corrected (`narrative_cohesion` always on roleplay-v1; `--per-turn` changes scoring, not the axis).

**Re-verified:** `pnpm -r typecheck` clean; eval 26, shared 59, engine 289, server non-e2e 356 green; `pnpm test:gate` passes (672 its).

yoarajota (approved):
lgtm

PR #19, opened 2026-08-20
fix(eval): local Qwen3 judge/generation calls leak <think> reasoning into JSON output

caiotheodoro:
## What
Two related `think:false`-suppression gaps on the local Ollama path:

- `packages/eval/src/run/scenario-runner.ts`: the eval harness's local-mode `LlmConfig` never set `extraBody.think: false` for the Ollama provider, unlike the product's own example config (`examples/simulations/qwen3-local.example.json`). Qwen3 emits a `<think>...</think>` reasoning block by default; without `think:false`, agent intent JSON never arrives and every turn falls back to a no-op.
- `packages/eval/src/judge/judge.ts`: the LLM judge's own calls hit Ollama's OpenAI-compatible `/v1/chat/completions` endpoint, which has no equivalent way to request `think:false` (only the native `/api/chat` path — see `ollama-provider.ts` — honors it). Fixed by stripping a `<think>...</think>` block before searching for the JSON object, and raising `max_tokens` so a thinking-mode model has room to reach its actual answer.

## Why
Both traced empirically to real benchmark runs, not theoretical:
- Generation: measured **100% fallback** before the fix (`No JSON object found in response` on nearly every turn), ~0% after.
- Judge: **12/16 scenarios' judge calls failed** in one real benchmark run with an opaque `Unexpected end of JSON input`, traced to the judge model spending its whole token budget on hidden reasoning.

## Verification
Re-ran the same benchmark scenarios after each fix — both failure rates dropped to ~0. `pnpm test:all` passes (764/764, no new tests needed — this package has no existing unit coverage over the Ollama request-body construction).

Stacked on #18 is unrelated (docs); this is the base of the narrative-sharpening fix stack — next PR builds on this one.

### DIFF
--- a/packages/eval/src/judge/judge.ts
+++ b/packages/eval/src/judge/judge.ts
@@ -31,6 +31,30 @@ function clampScore(v: number): number {
   return Math.min(5, Math.max(1, Math.round(v)));
 }
 
+/**
+ * Extracts the JSON object from a judge model's raw response text.
+ *
+ * The judge calls hit an OpenAI-compatible /chat/completions endpoint,
+ * which — unlike a native Ollama /api/chat call — has no way to request
+ * think:false for Qwen3-family models. Left uncontrolled, the model can
+ * spend its entire token budget on a <think>...</think> reasoning block and
+ * never emit the actual JSON, which used to blow up downstream as an opaque
+ * "Unexpected end of JSON input" on JSON.parse(""). Stripping the think
+ * block first (if present) means the brace search only ever looks at the
+ * model's actual answer.
+ */
+function extractJsonObject(raw: string): string {
+  const withoutThinking = raw.replace(/<think>[\s\S]*?<\/think>/gi, "");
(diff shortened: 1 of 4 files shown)


yoarajota (approved):

yoarajota:
did you checked if the real feature (not the eval) has this problem too?

caiotheodoro:
Good catch — yes, it did. `simulation-config.ts`'s `parseLlmConfig` passed `extraBody` through from `config/index.json` verbatim with no default; only the example configs set `extraBody.think: false` explicitly per agent, so any real config that didn't copy that field would hit the identical 100%-fallback bug in production, not just eval.

Pushed a fix on this branch (`0ecb831`): defaults `extraBody.think` to `false` when `providerType === "ollama"` and it's not explicitly set, leaves an explicit value untouched, and doesn't touch non-ollama providers. 3 new tests covering all three cases.

PR #20, opened 2026-08-20
fix(agent): intent validation false-rejections on create_channel/send_message

caiotheodoro:
## What
Two guard bugs in the intent-type-vs-available-action-allowlist check, plus two related normalizations, all found via live benchmark transcripts:

- **`personTargets` guard missing** in `intent-parser.ts` (present in `validate-intent.ts`, inconsistently): `send_message`'s available `personTargets` is always `[]` by design — any `personTargets` value on a `send_message` intent was rejected outright.
- **`channelTarget` guard missing in both enforcement sites** (`intent-parser.ts` **and** the engine's authoritative `validate-intent.ts`): `create_channel`'s available `channelTargets` is always `[]` — any `channelTarget` value on `create_channel` was rejected even at the authoritative layer.
- **`channelType: ""` default**: empirically the single most common `create_channel` failure — the model routinely emits an empty string instead of a valid enum value. Defaults to `"private_channel"` (matches the prompt's own guidance) instead of hard-rejecting. This fired on nearly every `create_channel` attempt across a 16-scenario benchmark run, invisible in fallback-count metrics since it's a validation rejection, not a provider failure.
- **Stray leading `#`** on `channelTarget`: the prompt's transcript rendering displays channels as `#channelId`; the model sometimes copies that display convention into the field value. Stripped before validation.
- **`action-intent-prompt-builder.ts`**: the JSON schema shown to the model listed one flat field set for every `intentType` and never mentioned `channelName`/`channelType`/`invitedAgentIds` (the fields `create_channel` actually needs) — very likely the root cause of the model guessing the wrong shape. Added per-`intentType` field notes.

## Verification
- New unit tests for the `channelType` default (empty string, missing, explicit-valid-value-untouched) — 3 new tests, 767/767 total passing.
- Re-ran the benchmark scenarios after this fix — the create_channel/send_message rejection errors present in nearly every prior transcript were gone.

Stacked on #19 (this branch's base) — that PR must merge first.

### DIFF
--- a/packages/server/src/agent/action-intent-prompt-builder.ts
+++ b/packages/server/src/agent/action-intent-prompt-builder.ts
@@ -29,16 +29,21 @@ JSON Schema:
   "id": "A unique string representing this intent (copy from the inputs or generate a new unique string)",
   "actorId": "${input.agentId}",
   "intentType": "Must match one of the available intent types",
-  "channelTarget": "Target channel ID if applicable (must match available targets)",
-  "personTargets": ["Array of target friend agentIds if applicable"],
+  "channelTarget": "ONLY for send_message/reply_to_message/leave_channel/invite_agent — the existing channel ID you're acting in. OMIT this field entirely for create_channel (use channelName/channelType instead).",
+  "personTargets": "OMIT this field for send_message and create_channel — it does not apply to them. Only used for actions that target a specific person directly.",
   "visibleContent": "Optional text message content (required for 'send_message' or 'reply_to_message')",
   "privateMotiveSummary": "A highly specific natural-language thought explaining your real, raw, hidden motive behind this action. Never leave empty.",
   "emotionDrivers": ["Emotion keywords driving this action (e.g. warmth, jealousy, irritation)"],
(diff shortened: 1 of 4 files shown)


packages/server/src/agent/action-intent-prompt-builder.ts
```diff
@@ -29,17 +29,22 @@ JSON Schema:
...
   "fallbackIfBlocked": "no_op",
-  "memoryWrites": []
+  "memoryWrites": [],
+  "channelName": "ONLY for create_channel — a short name for the new channel.",
+  "channelType": "ONLY for create_channel — usually \\"private_channel\\".",
+  "invitedAgentIds": ["ONLY for create_channel — array of agentIds to invite into the new channel."]
 }
 
+Field notes by intentType — send_message/reply_to_message use "channelTarget" (an existing channel); create_channel uses "channelName" + "channelType" (usually "private_channel") + "invitedAgentIds" instead, and has no "channelTarget" or "personTargets".
+
```
yoarajota, line 47: this file needs a full refactor... I'll implement and tag you

yoarajota (approved):

PR #22, opened 2026-08-20
feat(agent): structural repetition guard with regenerate-on-repeat retry

caiotheodoro:
## What
The prompt already tells the model not to repeat itself and (as of this stack) shows it the exact text to avoid, but empirically small local models repeat anyway — live benchmark runs showed every agent repeating its own line verbatim for 5-10 straight pulses despite the instruction. A prompt instruction alone isn't sufficient enforcement; this adds a structural backstop.

- **`PerceptionPacket.ownRecentUtterances`**: the agent's own last few sent messages, computed from the *full* visible event history in `build-perception-packet.ts` (before `CONTEXT_WINDOW` truncation) rather than the shared windowed view — in a 5-agent room the shared window fills up within 1-2 pulses, so an agent's own prior utterance can already be gone from context by the time it matters.
- **`repetition-guard.ts`**: `isNearRepeat()` — Jaccard word-overlap similarity (threshold 0.7, stopword-filtered) between a candidate message and the agent's recent utterances. Pure, unit-tested.
- **`action-intent-prompt-builder.ts`**: renders `ownRecentUtterances` as a concrete "you already said these, don't repeat them" list, and adds an instruction to actually use a relevant memory rather than let it sit unused (`memory_continuity` was one of the two weakest benchmark axes).
- **`agent-runtime.ts`**: on a detected near-repeat, retries generation *once* with an appended correction before falling back to `no_op` — a hard fallback on the first near-repeat just trades spam for silence; the retry recovers real content when the model has something else to say. Usage/token tracking sums both calls. Falls back to `no_op` (clearly labeled as post-retry) only if the retry also repeats or fails.

## Verification
- 6 new `repetition-guard` unit tests, 2 new `agent-runtime` tests (retry-recovers-content, retry-also-repeats-so-fallback). 775/775 total.
- Re-ran the 16-scenario benchmark sweep: **zero literal duplicate messages committed** (down from every agent repeating verbatim for 5-10 straight pulses); `narrative_cohesion` mean 3.00→3.80 and `memory_continuity` 2.93→3.67.

Stacked on #21 (this branch's base) — that PR must merge first. **Last PR in the narrative-sharpening stack** (chain: main ← #19 ← #20 ← #21 ← this one).

### DIFF
--- a/packages/server/src/agent/agent-runtime.ts
+++ b/packages/server/src/agent/agent-runtime.ts
@@ -11,6 +11,7 @@ import type {
 import { PersonaLoader } from "./persona-loader.js";
 import { PromptBuilder } from "./prompt-builder.js";
 import { IntentParser } from "./intent-parser.js";
+import { isNearRepeat } from "./repetition-guard.js";
 import { llmBudget } from "../llm/llm-budget.js";
 import { MockLlmProvider } from "../llm/mock-llm-provider.js";
 import { OpenAiCompatibleProvider } from "../llm/openai-compatible-provider.js";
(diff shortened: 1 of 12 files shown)


packages/server/src/agent/agent-runtime.ts
```diff
@@ -116,20 +117,73 @@ export class AgentRuntime {
...
+      (candidateIntent.intentType === "send_message" || candidateIntent.intentType === "reply_to_message") &&
+      !!candidateIntent.visibleContent &&
+      isNearRepeat(candidateIntent.visibleContent, input.perceptionPacket.ownRecentUtterances);
+
+    if (!fallbackApplied && isRepeat(intent)) {
+      repetitionRetried = true;
+      const retryPrompt = {
+        ...prompt,
+        system: `${prompt.system}\n\nIMPORTANT: your last attempt this turn ("${intent.visibleContent}") was too close to something you already said. Say something genuinely different — a new angle, a reaction to someone else, a topic change — or choose "no_op" if you truly have nothing new to add.`,
+      };
```
yoarajota, line 152: I'll implement a prompt builder ASAP

yoarajota (approved):

PR #52, opened 2026-08-21
feat(llm): seed Ollama generation for reproducible benchmark runs

matheusht:
## What

Adds deterministic sampling control to the LLM layer and the benchmark harness:

- `OllamaProvider` now translates a flat `extraBody.seed` into Ollama's nested `options.seed` (same pattern as the existing `top_p`/`repeat_penalty` translation). The OpenAI-compatible path already forwards `extraBody` at the body root, so `seed` works there for free.
- The eval harness's local-mode config pins a fixed seed (42, overridable via `PERFECTMAN_LLM_SEED`) into both provider branches, so two benchmark runs over the same scenarios are directly comparable instead of confounded by sampling variance. Normal simulation configs are untouched — they stay free to be nondeterministic.
- New test suites: fetch-mocked coverage for the Ollama provider (seed/repeat_penalty/top_p translation, absence when unset, no body-root leakage, URL derivation, conditional JSON format, result shape) and eval-side tests for the seed default/override/edge cases. Docs updated in `docs/eval/README.md`.

## Validation

- `pnpm lint` ✅ · `pnpm test:all` ✅ 792 passed
- Mock-mode bench slice runs green; blank/non-numeric seed overrides fall back to 42 while an explicit `0` stays respected.


matheusht:
Reviewer pass complete (independent subagent review): APPROVE verdict; both actionable findings were addressed in 7cf4657 — blank `PERFECTMAN_LLM_SEED` now falls back to 42 (explicit `0` still respected) with edge-case tests. Gates: `pnpm lint` ✅, `pnpm test:all` ✅ 792/792. Requesting review from @caiotheodoro and @yoarajota.

yoarajota:
@matheusht what is the usage of `seed`?

PR #53, opened 2026-08-21
fix(agent): reject content-bearing intents without visibleContent

matheusht:
## What

Closes a transcript-integrity hole in intent handling: an agent declaring `send_message` or `reply_to_message` with a missing or whitespace-only `visibleContent` used to pass both the intent parser and the engine validator (which only bounds max length), committing `message_sent`/`reply_sent` events with empty content into transcripts.

- The parser's post-validation guards now require non-blank `visibleContent` for content-bearing actions; violations route through the existing safe fallback ladder (clean `no_op` + descriptive error detail), exactly like every other malformed intent. Channel creation, reactions, and no-op intents are untouched.
- Updated one pre-existing test that had codified the old acceptance of contentless send intents; null-coercion coverage is preserved via a non-content action.
- Four new parser tests: missing content, whitespace-only content, reply intents, and the happy path.

Deliberately no change to the engine's type-generic validator — it remains the second gate.

## Validation

- `pnpm lint` ✅ · `pnpm test:all` ✅ 783 passed (778 baseline + net 5)

### DIFF
--- a/packages/server/src/agent/__tests__/intent-parser-null.test.ts
+++ b/packages/server/src/agent/__tests__/intent-parser-null.test.ts
@@ -12,7 +12,26 @@ const ACTIONS: AvailableAction[] = [
 ];
 
 describe("IntentParser structural repair", () => {
-  it("coerces visibleContent: null to undefined (model quirk)", () => {
+  it("coerces visibleContent: null to undefined (model quirk) on non-content actions", () => {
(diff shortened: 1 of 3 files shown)


yoarajota:
@matheusht no need to mention the issues in the code (like youre doing with #)

yoarajota (approved):

PR #55, opened 2026-08-21
docs(eval): refresh stale Calibration-pending section

matheusht:
## What

Refreshes the eval documentation's "Calibration-pending (honest list)" to match reality:

- The claim that no real local-model benchmark run existed is gone. The section now records that the run-blocking bugs were fixed, and cites the recorded sweep numbers: zero literal duplicate messages committed, `narrative_cohesion` 3.00→3.80, `memory_continuity` 2.93→3.67, aggregate signal pass 67.7%→70.8% on the 16-scenario slice.
- Judge calibration now cites the actual measured kappa (−0.116 vs the ≥0.7 gate) with its n=8 scope and the pulse-truncation confound, replacing the outdated "kappa = 0" framing.
- Names the two remaining trust gaps explicitly — unpinned sampling seed and same-family judge/generator — so readers know exactly why current numbers are directional.
- Every number traces to a committed artifact or recorded PR verification; nothing invented.

## Validation

Docs-only diff. `pnpm lint` ✅ · `pnpm test:all` ✅ 778 passed.

### DIFF
--- a/docs/eval/README.md
+++ b/docs/eval/README.md
@@ -80,8 +80,9 @@ turn is scored against the turn before it (see `packages/eval/src/judge/judge.ts
    scenario's full pulse count. The rule judge is v0 and the golden set
    still needs human review (Phase 2.4); the LLM judge must clear the
    kappa gate before its scores gate changes.
-3. **Real local-model data now exists.** The `think:false` / JSON-parse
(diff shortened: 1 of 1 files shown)


yoarajota:
Is there a reason to mention issues IDs in the codebase?

PR #58, opened 2026-08-21
ci: fast deterministic PR gate (typecheck, tests, golden-scenario bench)

matheusht:
## What

Introduces the repo's first CI: a fast, deterministic, fully offline PR gate.

- `.github/workflows/pr-gate.yml` runs on every PR and push to main: frozen-lockfile install → **full workspace build** (workspace packages resolve each other through `dist/`; skipping this produces phantom typecheck/test failures on a fresh checkout) → `pnpm lint` → `pnpm test:all` → a mock-generation + rule-judge bench over the 8 golden-labeled scenarios → hard gate assertion → report artifact.
- `scripts/ci/check-bench-gate.mjs` fails the run on any failed scenario or signal pass rate below 100% (the repo's iteration discipline), verifies every expected scenario id actually ran (variant-expanded `__vN` ids matched by base id), and exits 2 cleanly on malformed/missing reports. Rule-judge axis scores stay advisory by design until the judge clears its kappa calibration gate.
- The bench writes its report via an absolute `$GITHUB_WORKSPACE` path because `pnpm --filter` executes inside the package dir; assert/upload steps stay root-relative. Artifact upload is conditioned on report existence.
- Zero external dependencies: no model servers, no API keys. Doc pointer added in `docs/eval/README.md`.

## Validation

- Full pipeline reproduced locally exactly as CI invokes it: 26 variant runs across all 8 golden ids, 100% signal pass, `BENCH GATE PASSED`.
- Negative paths verified: mutated 90% report and missing-scenario-id reports both exit 1; malformed path exits 2.
- Workflow YAML validated; golden ids cross-checked against the scenario registry.

Issue #40, opened 2026-08-20
No CI-wired benchmark run — every run this session was manual/local


caiotheodoro:
Every benchmark run this session (calibration checks, the 16-scenario sweeps, the canary iteration loop) was kicked off manually against a local Ollama instance. There's no CI job that runs any of `packages/eval`'s bench/evidence tooling automatically, so a narrative-quality regression would currently only be caught if someone remembers to run it by hand.

A full LLM-judged run is too slow/expensive for per-PR CI (this session's 16-scenario sweep took ~35-45 minutes even on fast local hardware), but a lightweight periodic job (nightly/weekly) against a small canary scenario set (see the 4-scenario canary pattern used in PR #22's iteration loop) seems tractable, especially using the rule judge (`ruleJudge`, offline/deterministic, no LLM calls) as a fast per-PR smoke check with the LLM judge reserved for the periodic deeper run.

**Where to look**: `packages/eval/src/cli/bench.ts` / `packages/eval/src/cli/evidence.ts` (existing CLI entry points), `.github/workflows/` (if it exists) for the existing CI setup to extend.

matheusht:
## Plan

Add the recurring tier that complements the per-PR gate (which now exists):

1. `.github/workflows/benchmark.yml`: weekly cron + manual `workflow_dispatch`. Runs the same offline pipeline (build → lint → tests) then a FULL mock+rule-judge bench over all 123 rotated tasks — deterministic, free, no model servers — and uploads a date-stamped report artifact so quality trend over time becomes trackable instead of tribal memory.
2. `workflow_dispatch` carries an explicit opt-in input for future LLM-judged deep runs on self-hosted runners with models available; hosted runs stay offline by default. This keeps the "deeper LLM-judged narrative check" slot designed-in without wiring any live model into GitHub-hosted CI.
3. Gate assertion reused from `scripts/ci/check-bench-gate.mjs` (signals hold at 100%); rule-judge scores recorded in the artifact, advisory only.

**Validation**: workflow YAML validated; full mock bench + gate script executed locally end-to-end; dispatch-input logic reviewed for both branches (offline default / opt-in).

Closes #40 once shipped.

caiotheodoro:
Done by #59 — weekly full-suite benchmark workflow (benchmark.yml), llm dispatch gated to self-hosted runners.

PR #59, opened 2026-08-21
ci: weekly full-suite benchmark with trend artifacts

matheusht:
## What

Adds the recurring benchmark tier that complements the fast per-PR gate:

- `.github/workflows/benchmark.yml` runs weekly (Mondays 03:00 UTC) and on demand via `workflow_dispatch`: build → lint → tests → **full 123-task mock + rule-judge suite** (deterministic, offline, no model servers) → 100%-signals assertion → report artifact named per run, so quality trends become trackable over time instead of living in someone's session memory.
- The dispatch input exposes an explicit opt-in `judge=llm` mode for self-hosted runners that have model servers and keys configured — hosted CI stays offline by design; the deep LLM-judged narrative run gets a designed-in slot without wiring any live model into GitHub-hosted infrastructure.
- Reuses the shared bench-gate assertion script (this branch stacks on the PR-gate workflow branch; merge that one first).
- Report path uses an absolute workspace path (pnpm `--filter` executes inside the package dir); artifact upload is conditioned on report existence.

## Validation

- Full local reproduction: 123 scenarios, 0 failed, signal pass 100%, `BENCH GATE PASSED`.
- Workflow YAML validated; both dispatch branches (`rule` default / `llm` opt-in) reviewed.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -15,13 +15,18 @@ on:
           - rule
           - llm
 
+# A trend run should never lose a data point to a later dispatch, and the
+# scheduled cron shouldn't kill an in-flight opt-in deep run either — so
+# runs queue instead of cancelling each other.
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 jobs:
   benchmark:
-    runs-on: ubuntu-latest
+    # The offline path runs anywhere; the opt-in LLM deep run needs a runner
+    # with model servers + keys configured, so it targets self-hosted only.
+    runs-on: ${{ github.event.inputs.judge == 'llm' && 'self-hosted' || 'ubuntu-latest' }}
     # The opt-in LLM deep run needs far more headroom than the offline path.
     timeout-minutes: ${{ github.event.inputs.judge == 'llm' && 240 || 30 }}
     steps:
@@ -62,8 +67,9 @@ jobs:
 
       # Opt-in deep run (self-hosted runner with models configured via env).
       # Signal rates legitimately vary under real-model generation, so this
-      # path asserts only structural sanity: no scenario run hard-failed and
-      # every expected scenario id is present.
+      # path asserts only structural sanity: no scenario run hard-failed.
+      # (No expected-ids CSV is passed — the full suite has no fixed golden
+      # subset to check coverage against, unlike the PR-gate run.)
       - name: Deep bench (LLM judge)
         if: ${{ github.event.inputs.judge == 'llm' }}
         run: pnpm --filter @perfectman/eval bench --mode local --judge llm --out "$GITHUB_WORKSPACE/out/bench-full.json"
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -22,7 +22,8 @@ concurrency:
 jobs:
   benchmark:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    # The opt-in LLM deep run needs far more headroom than the offline path.
+    timeout-minutes: ${{ github.event.inputs.judge == 'llm' && 240 || 30 }}
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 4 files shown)


yoarajota:
🤔

caiotheodoro:
## Review: request changes — the `judge=llm` opt-in can never succeed as written

The true delta over #58 is small and mostly sound: valid YAML, correct weekly cron, schedule-triggered runs correctly fall through to the offline rule path, and the gate script's field contract matches `BenchReport`. One blocker:

### Blocking

**`judge=llm` dispatch is guaranteed to fail on every run.**
The workflow comments and PR body say llm mode "requires a runner with model servers + keys configured" / is "for self-hosted runners", but the job unconditionally uses `runs-on: ubuntu-latest` and sets no `PERFECTMAN_*` env or secrets. Dispatching `judge=llm` therefore runs `bench --mode local --judge llm` on a GitHub-hosted runner where nothing listens on `localhost:11434` (`judgeConfig()` defaults there when `PERFECTMAN_LLM_PROVIDER` is unset) — all 123 scenario runs throw, the gate fails on `scenariosFailed>0`, and the run burns hosted-runner time under the 240-minute timeout reserved for exactly this branch. Options: make `runs-on` conditional on the input (`${{ inputs.judge == 'llm' && 'self-hosted' || 'ubuntu-latest' }}`), add a fail-fast guard step that refuses `judge=llm` without a configured model endpoint, or drop the input until a self-hosted runner exists.

### Non-blocking

- `cancel-in-progress: true` on a shared concurrency group lets a manual dispatch cancel the scheduled trend run (losing that week's data point) and lets the Monday cron kill an in-flight 4-hour deep run. For a trend workflow, queue (`cancel-in-progress: false`) or include `github.event_name` in the group.
- **Artifacts are not a trend store**: default retention caps at 90 days and nothing aggregates them — "quality trends become trackable" degrades to a rolling window of disconnected JSONs someone must download and diff by hand. Also, `bench-${{ github.run_id }}` is unique but not date-stamped as the #40 plan promised. Standard fixes: `benchmark-action/github-action-benchmark` (gh-pages time series + regression alerts) or a committed rolling-baseline JSON.
- The comment above the deep-bench step claims the structural path asserts "every expected scenario id is present", but the assert step passes no expected-ids CSV, so that check is inert — only `scenariosFailed === 0` is enforced. Pass the id list or fix both comments.
- The seven-step setup sequence is duplicated verbatim from `pr-gate.yml`; since both first-ever workflows are landing together, extracting a composite action now is cheap and prevents drift.
- `docs/eval/README.md` documents only the PR gate — the weekly tier this PR adds is discoverable only by reading `.github/workflows/`.
- When llm mode does become real: a 100%-equality gate won't transfer to continuous judge scores — that tier needs an absolute floor + delta-vs-baseline (and judge self-agreement measurement), with scores advisory until the kappa calibration gate clears.

### Merge order

This branch is stacked directly on `ci/pr-gate-workflow` — the diff against main contains #58 verbatim. **#58 must merge first**, then this rebases to its true delta (benchmark.yml + the `--skip-signal-gate` extension of check-bench-gate.mjs).

---
*In-depth review by parallel agents; findings verified against the diff by execution, not description.*

yoarajota:
Addressed the blocking issue and two of the non-blocking ones:

- **`judge=llm` on hosted infra (blocking)**: fixed in 58161af. `runs-on` now resolves to `self-hosted` only when `judge=llm` is dispatched; the offline path still targets `ubuntu-latest`. A dispatch without a self-hosted runner configured now queues instead of running the full build/test/bench pipeline on hosted infra and failing at the deep-bench step.
- **`cancel-in-progress` on shared concurrency group**: fixed in 58161af — set to `false` so a manual dispatch can't drop a scheduled trend data point and the Monday cron can't kill an in-flight deep run.
- **Inert "every expected scenario id is present" comment**: fixed in 58161af — corrected the comment; no expected-ids CSV is passed on the full-suite run since there's no fixed golden subset to check coverage against there (unlike the PR-gate run).
- **Weekly tier undocumented**: fixed in 6d35f45 — added a section to `docs/eval/README.md`.

Left as residual, out of scope for this PR:
- Artifacts-as-trend-store (90-day retention, no aggregation) — agree this needs `benchmark-action/github-action-benchmark` or a committed rolling-baseline JSON; that's new tooling, tracking as follow-up rather than folding into this PR.
- Composite action to dedupe the pr-gate.yml/benchmark.yml setup steps — legitimate, but introducing a new reusable action is a structural change I'd rather get explicit sign-off on separately.
- Continuous judge-score gating design for when `llm` mode goes live — noted, deferred until that mode is actually exercised.

Local gate (`pnpm test:all`, 778 tests + hygiene audit) and CI both green as of 6d35f45.

matheusht:
@caiotheodoro friendly ping — both round-2 follow-ups are in (conditional deep-run timeout, structural-only assertion for the live path via `--skip-signal-gate`). Gates green locally and in CI when the branch last ran.

Issue #38, opened 2026-08-20
No constrained/schema-based decoding — still relying on format: "json" (syntax-only)


caiotheodoro:
`OllamaProvider` sets `body.format = "json"` (`packages/server/src/llm/ollama-provider.ts`), which only guarantees syntactically valid JSON — not that the shape matches `ActionIntentSchema`. Ollama (≥0.5) supports passing a full JSON-Schema object as `format` instead, which constrains generation to a specific field structure at the decoding level ("constrained decoding forces the model to emit only tokens that fit your schema, making malformed JSON mechanically impossible").

This session fixed several invalid-shape failure modes after the fact in validation (missing empty-allowlist guards, `channelType` defaulting — see PR #20) — schema-constrained decoding would prevent a class of these structurally, as defense-in-depth on top of the validation fixes, and could also make invalid `personTargets`/`channelTarget` combinations for a given `intentType` impossible to generate in the first place if the schema is made `intentType`-conditional (`oneOf`).

**Where to look**: `packages/server/src/llm/ollama-provider.ts` (`responseFormatJson`/`body.format`), `packages/shared/src/intent/intent.schema.ts` (`ActionIntentSchema`, the zod schema to convert — e.g. via `zod-to-json-schema`).

caiotheodoro:
Done by #65 (packet-schema constrained decoding on the Ollama native path) with #51's response_format support on the OpenAI-compatible path.

Issue #26, opened 2026-08-20
Novela scenario's judge call fails unpredictably — two distinct parse errors, unresolved


caiotheodoro:
The 90-pulse `novela_one_long_afternoon` scenario's LLM-judge scoring failed twice across this session's benchmark runs, with two *different* root causes:

1. Truncated reasoning — the judge spent its token budget on hidden `<think>` content and never reached the JSON answer (partially addressed by the `<think>`-stripping fix in PR #19, but not fully — see below).
2. The judge ignored the "Return ONLY a JSON object" instruction entirely and wrote a markdown-formatted prose critique instead ("**Roleplay Quality Score Summary** ... **1. Character Development (2/5)**..."), which `extractJsonObject` correctly rejected as unparseable rather than trying to salvage it.

Both failures happened specifically on the novela scenario (75-event transcript) and not on any of the 16 shorter scenarios in the same run — worth checking whether transcript length or event density is a factor (e.g. judge context getting crowded, or a validation error leaking into the transcript and confusing the judge — see linked issue on a leaked `channelType` error in transcripts).

**Where to look**: `packages/eval/src/judge/judge.ts`'s `llmJudge`/`extractJsonObject`, `packages/eval/src/cli/evidence.ts`'s `buildNovelaScenario`.

Originates from PR #22's benchmark run (`docs/eval/evidence/experiment-lowtemp-round2/`, locally — not currently committed).

matheusht:
## Plan

Two independent failure modes need two independent defenses:

1. **Truncated-reasoning runs (no JSON emitted at all)**: `llmJudge` retries ONCE with an escalated system instruction ("your previous reply was not parseable JSON; return ONLY the object") before giving up. Costs one extra call only when needed.
2. **Prose critiques that ignore the JSON contract**: new pure `salvageAxisScoresFromProse(raw, axisIds)` scans for rubric axis ids near 1–5 integer scores (handles `axis (2/5)`, `**axis**: 4`, `axis - 3` forms). Used strictly as a last resort after extraction+parse AND the retry both fail, and only accepted when ≥ half the rubric's axes are recovered (partial salvage beats a hard crash mid-benchmark; missing axes already default to 3 today).
3. Fixture-driven unit tests reproducing both observed failure shapes (prose critique with `(2/5)` scores; `<think>`-only truncation), plus fence-wrapped JSON, think-then-JSON, and below-threshold salvage rejection.

[...]

Issue #24, opened 2026-08-20
LLM judge fails calibration against golden labels (kappa = -0.116, target 0.7)


caiotheodoro:
The LLM-as-judge harness (`packages/eval/src/judge/judge.ts`) was run against the authored golden labels (`packages/eval/src/judge/golden-labels.ts`) using `calibrateJudge`, and failed badly: **kappa = -0.116** against a target of 0.7 (n=8 overlapping scenes). This means the judge's scores actively *disagree* with the project's own authored expectations of what good roleplay looks like — any benchmark number reported from this judge is directional at best, not trustworthy in absolute terms.

**Known confound worth investigating first**: the calibration run scored transcripts truncated at `pulseLimit=8`, while the golden labels were authored assuming each scenario's *full* pulse count plays out (14-36 pulses depending on scenario). That alone could explain a meaningful chunk of the disagreement — worth re-running calibration against full-length runs before concluding the judge itself is broken.

**Where to look**: `packages/eval/src/judge/calibration.ts` (the kappa/alpha computation, already built), `packages/eval/src/judge/golden-labels.ts` (the anchor set), `docs/eval/README.md`'s "Calibration-pending" section (already flags this as a known gap, item 2).

Originates from PR #22's benchmark run.

matheusht:
## Root-cause finding

The kappa=−0.116 run had one confound (pulseLimit=8 truncation), but inspection shows a second structural bug: variant-expanded bench runs register scenario ids as `<id>__vN`, while `calibrateJudge` looks up exact base ids — so on every rotated bench run the calibration matched ZERO scenes and reported a vacuous kappa 0 "FAIL". The harness has never actually measured judge-vs-golden agreement on those runs.

## Plan

1. New dedicated entry point `packages/eval/src/cli/calibrate.ts` (`pnpm --filter @perfectman/eval calibrate`): runs exactly the 8 golden-labeled scenarios at FULL pulse length (no truncation, no variant rotation — eliminating the documented confound structurally), scores them with `--judge rule|llm`, computes weighted kappa + alpha + disagreements, prints and writes the report. LLM mode stays available for maintainers' live reruns; this change ships validated offline with the rule judge.
2. Fix the vacuous match in `bench.ts`: calibration map keyed by base scenario id (suffix stripped) so rotated runs measure real agreement again.
[...]

PR #60, opened 2026-08-21
fix(eval): survive unparseable LLM judge responses with retry + prose salvage

matheusht:
## What

Makes LLM-judge scoring resilient to the two ways small judges actually break on long transcripts:

- **Strict retry**: when a response yields no parseable JSON, `llmJudge` retries once with an escalated system instruction ("respond ONLY with the JSON object — no prose, no markdown, no reasoning"). One extra call, only on failure.
- **Prose salvage as last resort**: new pure helper `salvageAxisScoresFromProse` recovers per-axis scores from markdown critiques (handles `axis (2/5)`, `axis: 4 out of 5`, `**axis**: 3` shapes), accepted only when ≥ half the rubric's axes are recovered; missing axes keep the existing default of 3. If nothing is recoverable, the thrown error stays honest and includes the raw prefix.
- Refactor: request/parse logic split into `callJudge`/`parseAxes` so the defense ladder reads linearly.
- 8 fixture-driven tests: fenced-JSON-with-prose first pass, think-truncation → retry success, double failure → salvage path, total failure → honest throw, plus salvage unit cases (threshold rejection, out-of-range scores, no-match).

## Validation

- \`pnpm lint\` ✅ · \`pnpm test:all\` ✅ 784 passed (778 baseline + net 6)
- All fixtures modeled on the real observed failure shapes recorded in the eval evidence notes (prose critique with `(N/5)` scores; `<think>` budget exhaustion).

### DIFF
--- a/packages/eval/src/cli/bench.ts
+++ b/packages/eval/src/cli/bench.ts
@@ -73,6 +73,7 @@ export type BenchReport = {
     axisScores: AxisScores;
     fallbackCount: number;
     latencyMs: number;
+    judgeSalvaged?: boolean;
     failed?: string;
   }>;
 };
@@ -137,14 +138,20 @@ export async function runBench(opts: {
       const artifact: ScenarioRunArtifact = await ScenarioRunner.run(scenario, { llmMode: mode });
       report.scenariosRun++;
 
-      const axisScores =
+      const judgeResult =
         judgeMode === "llm"
           ? perTurn
             ? await llmJudgePerTurn(scenario, artifact.events, judgeConfig())
             : await llmJudge(scenario, artifact.events, judgeConfig())
-          : ruleJudge(scenario, artifact.events, artifact.probeResults, artifact.passedSignals / Math.max(1, artifact.totalSignals));
+          : { axes: ruleJudge(scenario, artifact.events, artifact.probeResults, artifact.passedSignals / Math.max(1, artifact.totalSignals)), salvaged: false };
+      const axisScores = judgeResult.axes;
 
-      judgeScores.set(scenario.id, axisScores);
+      // A salvaged score is a fabricated/imputed read, not a clean parse —
+      // feeding it into calibration would silently compress the kappa the
+      // golden-label gate depends on.
+      if (!judgeResult.salvaged) {
+        judgeScores.set(scenario.id, axisScores);
+      }
       for (const [axis, v] of Object.entries(axisScores)) {
         axisAgg[axis] ??= { sum: 0, count: 0 };
         axisAgg[axis]!.sum += v;
@@ -176,6 +183,7 @@ export async function runBench(opts: {
         axisScores,
         fallbackCount: artifact.fallbackCount,
         latencyMs: artifact.latencyMs,
+        judgeSalvaged: judgeResult.salvaged,
       });
     } catch (err) {
       report.scenariosFailed++;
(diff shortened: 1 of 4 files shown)


caiotheodoro:
## Review: request changes — two blocking defects in the salvage layer (both verified by executing the regex)

The retry-then-salvage ladder is well structured and delivers the #26 plan, but the salvage layer needs fixes before this merges:

### Blocking

**1. The bare-digit branch fabricates scores from markdown list numbering.**
The regex alternative `(?=[.,;)\n]|$)` accepts any lone digit within 40 chars of an axis id. A numbered critique containing **zero actual scores** — e.g.

```
**1. in_character** — strong
**2. voice_match** — decent
**3. motive_authenticity** — shallow
**4. interpretation** ...
```

captures each *next* item's list number as the previous axis's score. Executing the exact regex: it salvages `in_character=2, voice_match=3, motive_authenticity=4, interpretation=5` — exactly 4 of 8 axes, which meets the ≥half threshold, so `llmJudge` returns a fully fabricated `AxisScores` (4 numbering artifacts + 4 defaulted 3s) instead of throwing honestly. Fix: require score-like context for the bare-digit branch (`/5`, `out of 5`) or anchor the match to the same line.

**2. Salvaged/imputed scores are indistinguishable from clean parses downstream.**
When salvage fires, up to half the axes are fabricated as 3 and the rest come from a lossy prose scan — but the return value carries no flag, no warning, no counter. `bench.ts` feeds it straight into `judgeAxisMeans` and `calibrateJudge(judgeScores, GOLDEN_LABELS)`, so the calibration kappa that #24 (and the new calibrate CLI) is trying to repair can be computed partly over invented midpoints with zero trace. Midpoint imputation also compresses variance, which alone can push weighted kappa toward or below zero. Minimum fix: log which axes were salvaged vs defaulted per scenario, propagate a `salvaged: true` marker into the report, and **exclude salvaged results from calibration** (or fail loudly there). For reference, this is where rigorous harnesses differ: lm-eval-harness returns an explicit `[invalid]` sentinel, OpenAI evals track `__invalid__` — nobody silently imputes.

### Non-blocking

- The one real-world failure quoted in #26 used judge-invented axis names (`**1. Character Development (2/5)**`) — the id-keyed scan (underscores → `[_\s]` only, no hyphens, no rubric labels) returns `null` on that exact shape, so the salvage mostly fires on shapes other than the one that motivated it. Consider also matching rubric labels, or temper the "modeled on the real observed failure shapes" claim.
- The salvage-fallback test asserts only `toBeDefined()`, which passes even if salvage output were ignored entirely (all axes default to 3). Assert the fixture's actual score, the fetch call count (2), and add a prose-then-garbage two-content fixture.
- `axis: 3.5/5` salvages as 3 (the `.` satisfies the punctuation lookahead) — fractional scores get floored, a systematic downward bias.
- `RETRY_SYSTEM_SUFFIX` says "your previous reply was not parseable JSON" but the retry is a fresh two-message conversation — the model never saw a previous reply. Replaying the failed output as an assistant turn would give it a concrete counterexample.
- Longer term: constrained decoding for the judge itself (Ollama `format` with the `{axes}` schema, per #38) would eliminate most of this failure class at the source; the salvage ladder is a sound last-resort layer but shouldn't be the only one.

### Merge order

`judge.ts` is also rewritten by #51 (PromptSection prompts, `LlmJudgeConfig` → `LLMJudgeConfig`, which keeps the OLD single-attempt `llmJudge` body). Land this PR first, then rebase #51 — a naive resolution taking #51's `llmJudge` silently reverts the entire retry/salvage ladder.

---
*In-depth review by parallel agents; findings verified against the diff by execution, not description.*

yoarajota (approved):

yoarajota:
on it @caiotheodoro @matheusht

PR #61, opened 2026-08-21
feat(eval): full-length judge calibration CLI + real bench calibration

matheusht:
## What

Makes judge-vs-golden agreement a measurement that actually happens, instead of two stacked artifacts:

- **New `calibrate` CLI** (`pnpm --filter @perfectman/eval calibrate --judge rule|llm --out ...`): plays each golden-labeled scenario at its **full pulse count** — no truncation, no variant rotation — scores it with the rule or LLM judge, and reports weighted kappa, alpha, per-axis kappa, and concrete disagreements. This eliminates the truncated-transcript confound structurally: the transcripts a judge is calibrated on are exactly the shape the golden labels were authored for. LLM mode stays available for maintainers' live-model reruns; this ships validated offline via the rule judge.
- **Real bench calibration**: rotated bench runs register scenario ids as `<id>__vN`, while `calibrateJudge` looks up exact base ids — so every expanded run to date matched **zero** golden scenes and reported a vacuous kappa 0 "FAIL". `bench.ts` now folds variant ids back to base ids for the calibration map (`baseScenarioId()` helper), so those runs measure genuine agreement.
- Honest offline baseline committed as evidence (`docs/eval/evidence/calibration-mock-full.json`) — mock+rule judge FAILs the 0.7 gate as expected for a proxy; committed anyway per repo discipline.
- Unit tests: base-id stripping, base-keyed matching (identical raters → kappa 1 across scenes), vacuous-empty case.

## Validation

- \`pnpm lint\` ✅ · \`pnpm test:all\` ✅ 780 passed
- Live CLI run over all 8 golden scenes at full length produced real per-axis kappas + disagreement lists (e.g. rule-judge voice_match reads 2 where goldens expect 4) — exactly the actionable signal calibration was supposed to provide.


caiotheodoro (approved):
LGTM

Issue #50, opened 2026-08-20
Wire fallbackIfBlocked: honor the model-declared alternate action when the primary is denied (single-level, engine-clamped)


yoarajota:
## Problem

`fallbackIfBlocked` is a model-sourced decision ("if this action is denied, fall back to X") that, until recently, was **dead**: the resolver returned `"blocked"` and stopped, never honoring the declared alternate. It was flagged as dead during the #49 planning review, and a first wiring landed in `packages/server/src/simulation/intent-resolver.ts` — but the design has real open questions that deserve their own issue rather than being folded into the prompt-refactor work (#49).

This matters because a fallback is *not* just "turn into no_op" — it can trigger a **new sequence of events** (e.g. primary `create_channel` blocked → fallback `send_message` in the current pulse). The engine should honor that decision, correctly and safely.

## Current implementation (landed with #49 hardening, reference only)

`intent-resolver.ts`:
- `resolve()` wraps `resolveInternal()`; if the primary is `"blocked"` and `intent.fallbackIfBlocked` is set, it builds a derived alternate intent via `buildFallback()` and re-resolves it (single level — the derived intent's `fallbackIfBlocked` is cleared, so no recursion).
- Returns the existing `"fallback_committed"` outcome when the fallback commits/delays; returns `"blocked"` (with both block events) if the fallback is also blocked.
- The primary `intent_blocked` event is retained so telemetry shows the denial.
- Supported fallback targets: `no_op`, `delay_response`, `send_message`, `reply_to_message`, `react`, `create_channel`, `leave_channel`, `invite_agent`. `typing_*`/`write_memory` return `null` (not meaningful) → plain block.
- Tests: model-provided channelType registers real type; fallback commits alternate + keeps block event; still-blocked fallback returns `blocked`. Server suite green (415).

## Open design questions (need decisions)

1. **Depth / sequencing.** Current fallback is single-level and same-pulse. Do we ever want multi-pulse or multi-level fallback chains ("if send_message denied, wait and then try a private channel")? If so, define the sequencing model and a depth cap.
2. **Target-type allow-list.** Is the current set right? Should `create_channel`/`invite_agent` as fallbacks be allowed, or restricted to low-risk surfaces (`no_op`/`delay_response`/`send_message`/`reply`/`react`)?
[...]

PR #64, opened 2026-08-21
feat(simulation): honor model-declared fallbackIfBlocked under engine clamps

matheusht:
## What

Makes `fallbackIfBlocked` a real, safe mechanism instead of dead schema: when an agent's primary action is denied, its declared alternate now resolves through the same pipeline — once, same pulse, under engine control.

Policy decisions encoded (and tested):

- **Engine clamp**: only low-risk surfaces are reachable as fallbacks — `no_op`, `delay_response`, `send_message`, `reply_to_message`, `react`. Side-effect-heavy types (`create_channel`, `invite_agent`, `write_memory`) can never be reached via a denied intent.
- **Rate-limit anti-gaming**: a primary blocked by the rate-limit gate suppresses fallback entirely; validation/membership/blocked-action denials honor it.
- **No chains by construction**: the derived intent receives a fresh id and its own `fallbackIfBlocked` is cleared.
- **No empty-message loophole**: content-bearing fallbacks require non-blank `visibleContent`, else the original denial stands.
- **Telemetry preserved**: the primary's `intent_blocked` event ships inside the fallback result; success surfaces as the existing-but-unused `"fallback_committed"` outcome.
- The derived intent re-enters full resolution (validation, membership, gates) — model proposes, engine approves.

## Validation

- \`pnpm lint\` ✅ · \`pnpm test:all\` ✅ 812 passed (+6 dedicated tests)
- Covers: commit-after-denial with retained block telemetry, rate-limit suppression, clamp rejection, no-chaining event count, empty-content guard, illegal-target re-blocking.

### DIFF
--- a/packages/server/src/simulation/__tests__/intent-resolver-fallback.test.ts
+++ b/packages/server/src/simulation/__tests__/intent-resolver-fallback.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, beforeEach } from "vitest";
 import { IntentResolver } from "../intent-resolver.js";
-import type { RateLimitGate } from "../rate-limit-gate.js";
+import { RateLimitGate } from "../rate-limit-gate.js";
 import { ChannelRegistry } from "../channel-registry.js";
 import { InMemoryChannelRepository } from "../in-memory-stores.js";
 import type {
@@ -104,11 +104,13 @@ describe("IntentResolver fallbackIfBlocked (#50 policy)", () => {
       updatedAt: Date.now(),
     });
     await channelRepo.addMembership({ channelId: CHANNEL_ID, agentId: AGENT_ID, joinedAt: Date.now() });
-    const rateLimitGate = new RateLimitGateStub();
-    resolver = new IntentResolver(rateLimitGate as RateLimitGate, channelRegistry);
+    resolver = new IntentResolver(new RateLimitGate(SETTINGS), channelRegistry);
   });
 
-  const ctx = (availableActions = AVAILABLE_ACTIONS) => ({
+  const ctx = (
+    availableActions = AVAILABLE_ACTIONS,
+    settings: SimulationSettings = SETTINGS,
+  ) => ({
     simulationId: SIM_ID,
     channelId: CHANNEL_ID,
     pulseIndex: 1,
@@ -116,7 +118,7 @@ describe("IntentResolver fallbackIfBlocked (#50 policy)", () => {
     availableActions,
     channels: [],
     membership: [],
-    settings: SETTINGS,
+    settings,
     actionEmotions: ACTION_EMOTIONS,
   });
 
@@ -131,18 +133,22 @@ describe("IntentResolver fallbackIfBlocked (#50 policy)", () => {
   });
 
   it("suppresses fallback entirely when the primary was rate-limited", async () => {
-    const denyingGate = {
-      allowAction: () => false,
-      recordAction: () => {},
+    // A zero message limit makes the real gate deny send_message — no stub.
+    const zeroMessageSettings: SimulationSettings = {
+      ...SETTINGS,
+      maxMessagesPerMinutePerAgent: 0,
     };
-    const gated = new IntentResolver(denyingGate as unknown as RateLimitGate, channelRegistry);
+    const gated = new IntentResolver(new RateLimitGate(zeroMessageSettings), channelRegistry);
 
     const intent = makeIntent({ fallbackIfBlocked: "no_op" });
-    const result = await gated.resolve(intent, ctx());
+    const result = await gated.resolve(intent, ctx(AVAILABLE_ACTIONS, zeroMessageSettings));
 
     expect(result.outcome).toBe("blocked");
     expect(result.committedEvents).toHaveLength(1);
     expect(result.committedEvents[0]!.type).toBe("intent_blocked");
+    const violations = (result.committedEvents[0]!.payload as { violations: Array<{ type: string }> })
+      .violations;
+    expect(violations[0]!.type).toBe("rate_limited");
   });
 
   it("rejects side-effect-heavy fallback types (engine clamp)", async () => {
@@ -207,11 +213,3 @@ describe("IntentResolver fallbackIfBlocked (#50 policy)", () => {
     expect(result.committedEvents).toHaveLength(1);
   });
 });
-
-class RateLimitGateStub {
-  allowed = true;
-  async allowAction(): Promise<boolean> {
-    return this.allowed;
-  }
-  recordAction(): void {}
-}
--- a/packages/server/src/simulation/__tests__/intent-resolver-fallback.test.ts
+++ b/packages/server/src/simulation/__tests__/intent-resolver-fallback.test.ts
@@ -1,0 +1,207 @@
+import { describe, it, expect, beforeEach } from "vitest";
+import { IntentResolver } from "../intent-resolver.js";
+import type { RateLimitGate } from "../rate-limit-gate.js";
+import { ChannelRegistry } from "../channel-registry.js";
+import { InMemoryChannelRepository } from "../in-memory-stores.js";
+import type {
+  ActionIntent,
+  AgentState,
+  SimulationSettings,
+  ActionEmotions,
+  AvailableAction,
+} from "@perfectman/shared";
+import { createId } from "@perfectman/shared";
+
+const SIM_ID = "sim_fb";
+const AGENT_ID = "agent_1";
+const CHANNEL_ID = "ch_public";
+
+const SETTINGS: SimulationSettings = {
+  omniscientSpectatorMode: false,
+  allowPrivateChannels: true,
+  maxPrivateChannelsPerAgent: 3,
+  maxMessagesPerMinutePerAgent: 20,
+  llmCallBudgetPerMinute: 10,
+  pulseIntervalMs: 3000,
+  tokenBudgetPerHour: 1_000_000,
+};
+
+function makeIntent(overrides: Partial<ActionIntent> = {}): ActionIntent {
+  return {
+    id: createId(),
+    actorId: AGENT_ID,
+    intentType: "send_message",
+    channelTarget: CHANNEL_ID,
+    personTargets: [],
+    visibleContent: "mensagem de verdade",
+    privateMotiveSummary: "test motive",
+    emotionDrivers: [],
+    motivationDrivers: [],
+    memoryWrites: [],
+    ...overrides,
+  };
+}
+
+function makeAgentState(): AgentState {
+  return {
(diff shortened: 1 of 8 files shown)


packages/server/src/simulation/intent-resolver.ts
```diff
@@ -289,16 +303,108 @@ function buildNoOpEvent(
...
+   *   events of a successful fallback; the outcome becomes the existing
+   *   "fallback_committed".
+   */
   async resolve(
     intent: ActionIntent,
     ctx: ResolveContext,
   ): Promise<ResolvedIntent> {
+    const primary = await this.resolveInternal(intent, ctx);
+    if (primary.result.outcome !== "blocked") return primary.result;
+    if (primary.rateLimited) return primary.result;
```
caiotheodoro, line 396: bug: the rate-limit suppression only fires on the in-resolver `RateLimitGate` path, but in production rate limits are baked into `availableActions` before the LLM call (pulse-scheduler.ts:198 → `computeAvailableActions`), so a rate-limited primary arrives as a `validateIntentPure` violation with `rateLimited=false` — and the fallback IS honored. The invariant the PR body claims is inverted on the real path; the flag needs to also treat validation blocks whose `blockReason` is rate-limit-derived (`message_rate_limit`, `messages_per_minute_exceeded`) as suppressing.

packages/server/src/simulation/intent-resolver.ts
```diff
@@ -289,16 +303,108 @@ function buildNoOpEvent(
...
+  if (fallbackType === "send_message" || fallbackType === "reply_to_message") {
+    derived.visibleContent = primary.visibleContent;
+    derived.channelTarget = primary.channelTarget;
+  }
+  if (fallbackType === "reply_to_message") {
+    derived.replyToEventId = primary.replyToEventId;
+  }
+  if (fallbackType === "react") {
+    derived.targetEventId = primary.targetEventId;
+    derived.emoji = primary.emoji;
```
caiotheodoro, line 361: risk: the react branch drops `channelTarget`, so the derived intent skips both the hidden_channel_target validation and the `isMember` check (which only runs `if (intent.channelTarget)`) and the reaction re-homes to `ctx.channelId` — a react blocked for targeting a channel the agent isn't in can re-commit via fallback with the same `targetEventId`, referencing an event the agent can't see.
```suggestion
    derived.emoji = primary.emoji;
    derived.channelTarget = primary.channelTarget;
```

packages/server/src/simulation/intent-resolver.ts
```diff
@@ -25,6 +25,20 @@ export type ResolvedIntent = {
...
 
+/**
+ * Engine clamp for model-declared `fallbackIfBlocked`: only low-risk
+ * surfaces are reachable when the primary action is denied. Side-effect-
+ * heavy types (create_channel, invite_agent, write_memory) are excluded —
+ * a denied intent must not become a route to structural changes.
+ */
+const FALLBACK_ALLOWED_TYPES: ReadonlySet<ActionIntent["intentType"]> = new Set([
+  "no_op",
+  "delay_response",
```
caiotheodoro: risk: `delay_response` is in the allow-list but `deriveFallbackIntent` has no branch for it and never carries `preferredDelay`, and `intentToEvents` maps it to `buildNoOpEvent` — so this fallback always commits an immediate no_op whose payload claims `delay_response`. It's also the only allow-list member with zero test coverage. Remove it from the set or document and test it as a deliberate no_op alias.

packages/server/src/simulation/__tests__/intent-resolver-fallback.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    const gated = new IntentResolver(new RateLimitGate(zeroMessageSettings), channelRegistry);
+
+    const intent = makeIntent({ fallbackIfBlocked: "no_op" });
+    const result = await gated.resolve(intent, ctx(AVAILABLE_ACTIONS, zeroMessageSettings));
+
+    expect(result.outcome).toBe("blocked");
+    expect(result.committedEvents).toHaveLength(1);
+    expect(result.committedEvents[0]!.type).toBe("intent_blocked");
+    const violations = (result.committedEvents[0]!.payload as { violations: Array<{ type: string }> })
+      .violations;
```
caiotheodoro, line 154: risk: this suppression test constructs a state production can't reach — `availableActions` hand-built with send_message unblocked while the gate limit is 0, but `computeAvailableActions` derives from the same counters, so the real pipeline blocks at validation and the fallback commits, the opposite of what's pinned here. Add a companion test feeding `availableActions` marked blocked with `blockReason: "message_rate_limit"` asserting suppression — it will fail against the current code, which is the point.

packages/server/src/simulation/intent-resolver.ts
```diff
@@ -289,16 +303,108 @@ function buildNoOpEvent(
...
+      return primary.result;
+    }
+    if (fallbackType === "react" && !intent.targetEventId) {
+      // Same orphan-reference rule for reactions.
+      return primary.result;
+    }
+
+    const derived = deriveFallbackIntent(intent, fallbackType);
+    const secondary = await this.resolveInternal(derived, ctx);
+    if (secondary.result.outcome !== "committed") {
```
caiotheodoro, line 421: question: issue #50's reference implementation returned both block events when the fallback was also denied; here the secondary's block event is discarded, so operators get no trace a fallback was attempted and re-blocked. Intentional simplification, or should the derived intent's block event ride along in `committedEvents`?

packages/server/src/simulation/__tests__/intent-resolver-fallback.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    expect(result.committedEvents.map(e => e.type)).toEqual(["intent_blocked"]);
+  });
+
+  it("does not chain: a second-level declaration on the derived intent is impossible", async () => {
+    // Primary send_message blocked; fallback no_op commits. Even though the
+    // derived intent spreads from the primary (which declared a fallback),
+    // the resolver clears it — observable via exactly one extra event and
+    // no repeated resolution attempts.
+    const intent = makeIntent({
+      fallbackIfBlocked: "no_op",
```
caiotheodoro, line 227: nit: the name claims chaining is impossible, but the fallback commits on the first try so no chain is attempted — what it actually pins is the memoryWrites drop (good test, mislabeled). Chaining is structurally unobservable anyway since `resolveInternal` never consults `fallbackIfBlocked`; rename to what it verifies.

caiotheodoro (commented):
One bug and three risks inline. The clamp list, single-level guarantee, and retained intent_blocked telemetry all check out. PR-wide: this collides with #51, whose packet schema excludes `fallbackIfBlocked` (additionalProperties:false, prompt says "never set them", parser test asserts it's undefined) — if #51 merges, this entire fallback branch is dead code for every LLM-sourced intent. One contract has to change explicitly: #51 re-admits the field as model-decided, or #64 gets an engine-side fallback policy. Both PRs also edit intent-resolver.ts, so the second one in needs a semantic re-review.

yoarajota:
went with option B — engine-side policy instead of putting fallbackIfBlocked back on the model packet. reason: preferredDelay is already the precedent for a structural field the engine owns (composeIntentPacket always stamps it, model never gets a say), and action-intent-prompt-builder.ts already tells the model fallbackIfBlocked is system-assigned, so re-admitting it to ModelIntentPacketSchema wouldve contradicted a contract that already exists post-#51. composeIntentPacket now derives it itself from intentType — send_message/reply_to_message/react/create_channel/invite_agent get an automatic no_op fallback, everything else doesnt need one.

fixed the other three bugs too: rate-limit suppression now checks the validation violation detail (message_rate_limit / messages_per_minute_exceeded) not just the in-resolver gate, so it actually suppresses on the real computeAvailableActions path; react fallback carries channelTarget now so it re-validates against the real target instead of re-homing to ctx.channelId; dropped delay_response from the allow-list instead of wiring a real delay — carrying preferredDelay through wouldve meant resolve() treating a delayed secondary as a valid outcome, more than this pr should take on. also went with retaining the secondarys block event when the fallback is denied too (matches how the primarys block event already rides along on a successful fallback), and renamed the mislabeled "no chains" test to what it actually pins.

all covered by new/updated tests, pnpm test:all green (875 tests, gate pass), typecheck clean on all 4 packages, ci green. ready for a look.

packages/server/src/simulation/intent-resolver.ts
yoarajota: good catch, fixed — resolveInternal now flags rateLimited when the block violation detail is message_rate_limit or messages_per_minute_exceeded too, not just the in-resolver gate check. added a test that builds availableActions through computeAvailableActions so it actually exercises the real path instead of a state production cant reach

packages/server/src/simulation/intent-resolver.ts
yoarajota: nice catch, took your suggestion basically as-is — deriveFallbackIntent carries channelTarget for react now, added a test pinning that it re-blocks against the real target instead of re-homing to ctx.channelId

packages/server/src/simulation/intent-resolver.ts
yoarajota: dropped it from the allow-list instead of wiring a real delay — carrying preferredDelay through wouldve meant resolve() treating a delayed secondary as a valid outcome, thats more infra than this pr should take on. added a test pinning the rejection

packages/server/src/simulation/__tests__/intent-resolver-fallback.test.ts
yoarajota: rewrote it to build availableActions through computeAvailableActions with maxMessagesPerMinutePerAgent 0, so it actually blocks at validateIntentPure with message_rate_limit like production does. kept the old RateLimitGate-only test too since thats still a real defense-in-depth path, just renamed it

packages/server/src/simulation/intent-resolver.ts
yoarajota: went with retaining it — resolve() already keeps the primary block event when the fallback succeeds, so now the secondarys block event rides along the same way when its denied too. outcome stays blocked, committedEvents has both

packages/server/src/simulation/__tests__/intent-resolver-fallback.test.ts
yoarajota: renamed to "drops the denied primary's memoryWrites from the committed fallback", thats what it actually pins

matheusht:
@caiotheodoro ping — both medium policy gaps fixed (per-type fallback derivation so no orphan refs, denied primaries never commit their memoryWrites); the engine-side violation-type rename is documented as a separate decision on the issue thread.

PR #65, opened 2026-08-21
feat(llm): schema-constrained intent decoding for the Ollama path

matheusht:
## What

Adds decoding-level shape enforcement for local model intents:

- `ACTION_INTENT_JSON_SCHEMA`: derived from the zod `ActionIntentSchema` via zod-to-json-schema, inlined flat (no `$ref` wrappers). The zod schema remains the single source of truth — no second hand-written contract to drift.
- New `constrainedDecoding` flag on `LlmConfig` (off by default, Ollama native path only): when enabled together with `responseFormatJson`, the request's `format` carries the full JSON Schema instead of bare "json", so tokens violating the intent shape become mechanically impossible at decode time (Ollama >= 0.5).
- IntentParser's repair ladder stays untouched — constrained decoding is defense in depth for the flagged path, not a replacement for validation elsewhere.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS 812 passed (+8 new)
- Provider tests assert the schema object lands in `format` (intentType enum + field shapes), stays "json" without the flag, and stays absent when JSON mode is off; shared-side tests pin flatness/no-refs, enum completeness, required fields, and sync with a canonical intent.

### DIFF
--- a/docs/test-inventory.json
+++ b/docs/test-inventory.json
@@ -1,8 +1,8 @@
 {
-  "generatedAt": "2026-08-22T06:52:31.956Z",
+  "generatedAt": "2026-08-22T07:08:50.528Z",
   "totals": {
     "files": 86,
-    "its": 783
+    "its": 785
   },
   "byPackage": {
     "engine": {
@@ -25,7 +25,7 @@
     },
     "shared": {
       "files": 7,
-      "its": 71,
+      "its": 73,
       "flags": []
     }
   },
@@ -1820,7 +1820,7 @@
       "file": "packages/shared/src/__tests__/intent-packet-schema.test.ts",
       "pkg": "shared",
       "layer": "contract",
-      "its": 7,
+      "its": 9,
       "itsEach": 0,
       "forcedCasts": 0,
       "tsBypasses": 0,
--- a/docs/test-inventory.md
+++ b/docs/test-inventory.md
@@ -1,6 +1,6 @@
 # Test Inventory & Hygiene Snapshot
 
-Generated: 2026-08-22T06:52:31.950Z — audited 86 files, 783 `it`/`test` blocks (some it.each expand further at runtime).
+Generated: 2026-08-22T07:08:50.523Z — audited 86 files, 785 `it`/`test` blocks (some it.each expand further at runtime).
 
 Layer classification follows `docs/testing-strategy.md`: `contract` (zod/boundary, once per package) / `integration` (through public surface, the honeycomb's big middle) / `e2e` (max 6 suites) / `eval-harness` (vitest tests of the eval tooling — the 123-task *benchmark* is out of scope).
 
@@ -11,9 +11,9 @@ Layer classification follows `docs/testing-strategy.md`: `contract` (zod/boundar
 | engine | 21 | 246 | —
 | eval | 9 | 59 | —
 | server | 49 | 407 | console(1), async-pause(1)
-| shared | 7 | 71 | —
+| shared | 7 | 73 | —
 
-**Total: 86 files, 783 it/test blocks; 2 files flagged.**
+**Total: 86 files, 785 it/test blocks; 2 files flagged.**
 
 ## Per-file inventory
 
@@ -100,7 +100,7 @@ Layer classification follows `docs/testing-strategy.md`: `contract` (zod/boundar
 | packages/server/src/simulation/__tests__/world-signals-builder.test.ts | integration | 7 | 
 | packages/shared/src/__tests__/constants.test.ts | contract | 13 | 
 | packages/shared/src/__tests__/contracts.test.ts | contract | 3 | 
-| packages/shared/src/__tests__/intent-packet-schema.test.ts | contract | 7 | 
+| packages/shared/src/__tests__/intent-packet-schema.test.ts | contract | 9 | 
 | packages/shared/src/__tests__/prompt-syntax.test.ts | contract | 5 | 
 | packages/shared/src/__tests__/schemas.test.ts | contract | 22 | 
 | packages/shared/src/persona-packs/persona-packs.test.ts | contract | 13 | 
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -22,7 +22,8 @@
     "vitest": "^2.0.0"
   },
   "dependencies": {
-    "zod": "^3.23.0",
+    "zod": "^3.25.28",
+    "zod-to-json-schema": "^3.25.2",
     "nanoid": "^5.0.0"
   }
 }
--- a/packages/shared/src/__tests__/intent-packet-schema.test.ts
+++ b/packages/shared/src/__tests__/intent-packet-schema.test.ts
@@ -6,7 +6,7 @@ import {
   composeIntentPacket,
   memoryWriteProposalFieldContract,
 } from "../intent/intent-packet.schema.js";
-import { MemoryWriteProposalSchema } from "../intent/intent.schema.js";
+import { MemoryWriteProposalSchema, IntentTypeSchema } from "../intent/intent.schema.js";
 
 describe("ModelIntentPacket", () => {
   it("excludes engine-stamped structural fields (id, actorId, preferredDelay, fallbackIfBlocked)", () => {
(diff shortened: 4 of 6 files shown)


packages/shared/src/scenario/scenario.types.ts
```diff
@@ -0,0 +1,46 @@
+export type ScenarioPreset =
```
caiotheodoro: bug: this file plus `scenario.schema.ts` and `scenario-presets.ts` (184 lines) landed in 50e5b8c, whose message never mentions them. They're unrelated to constrained decoding, unexported from the package index, and untested — dead code from another workstream. Drop them from this branch and land them in their own PR.

packages/server/src/llm/ollama-provider.ts
```diff
@@ -86,7 +87,14 @@ export class OllamaProvider implements LlmProvider {
     }
 
     if (this.config.responseFormatJson) {
-      body.format = "json";
+      // Constrained decoding: when enabled, the schema itself becomes the
+      // format target (Ollama >= 0.5) so STRUCTURALLY invalid JSON becomes
+      // impossible at decode time. Note Ollama's grammar enforces
+      // shape/types/enums but not value constraints (minLength, minimum) —
+      // IntentParser's repair ladder remains the backstop everywhere.
+      body.format = this.config.constrainedDecoding
```
caiotheodoro: risk: Ollama ~0.30.0–0.31.1 silently drops `format` (HTTP 200, no error) when `think:false` is set on qwen3-family models (ollama#14645, ollama#15260) — and `think:false` is what this provider sends by default since parseExtraBody's defaulting. On those versions the "mechanically impossible" comment is silently false. Document the real minimum version (post-#15260, not ">= 0.5") and emit a counter when `constrainedDecoding` was on but `IntentParser` still had to repair structure — that's the tell the grammar was dropped.

packages/eval/src/run/scenario-runner.ts
```diff
@@ -312,6 +312,7 @@ export function localLlmConfig(pack: import("@perfectman/shared").PersonaPack |
     timeoutMs: 120000,
     retryCount: 2,
     responseFormatJson: true,
+    constrainedDecoding: !isDeepseek,
```
caiotheodoro: risk: this default-enables constrained decoding for every local eval run, contradicting the PR body's off-by-default framing, in the one config combining every failure mode: qwen3 + `think:false` (~15 lines down) + `seed` — format-plus-seed is reported non-deterministic (ollama#12559), so this can undo #52's reproducibility. It also breaks comparability with every prior bench in docs/eval/evidence/. Gate behind an env var defaulting off, or at minimum record the flag in bench metadata.

packages/shared/src/intent/intent-json-schema.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * JSON Schema derived from ActionIntentSchema — the single source of truth
+ * (see #49's drift concern: hand-written contract copies are how prompt and
+ * validation drifted apart historically).
+ *
+ * Consumed by OllamaProvider as a structured-output `format` target, so the
+ * model can only emit tokens that fit the intent shape at decoding time.
+ */
+export const ACTION_INTENT_JSON_SCHEMA: Readonly<Record<string, unknown>> = Object.freeze(
+  zodToJsonSchema(ActionIntentSchema, {
```
caiotheodoro: risk: the derived `required` is [id, actorId, intentType, privateMotiveSummary], so the grammar forces the model to invent `id` and echo `actorId` — and `IntentParser` never verifies a model-supplied `actorId`, which the resolver then trusts for event attribution. Derive the format target from a model-decision packet that omits engine-stamped fields (converging with #51), or pin `actorId` to a per-request const in the schema so the grammar can't attribute the intent to another agent.

packages/shared/src/__tests__/intent-json-schema.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  it("is a flat object schema with no dangling refs", () => {
+    expect(schema.type).toBe("object");
+    expect(JSON.stringify(schema)).not.toContain("$ref");
+    expect(schema.additionalProperties).toBe(false);
+  });
+
+  it("exposes the full intent-type enum", () => {
+    expect(schema.properties.intentType?.enum).toContain("send_message");
```
caiotheodoro: nit: "enum completeness" checks 2 of 11 intent types by containment — removing the other 9 from zod passes silently. `expect(schema.properties.intentType?.enum).toEqual(IntentTypeSchema.options)` pins the whole set.

packages/shared/src/intent/intent-json-schema.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * JSON Schema derived from ActionIntentSchema — the single source of truth
+ * (see #49's drift concern: hand-written contract copies are how prompt and
+ * validation drifted apart historically).
+ *
+ * Consumed by OllamaProvider as a structured-output `format` target, so the
+ * model can only emit tokens that fit the intent shape at decoding time.
+ */
+export const ACTION_INTENT_JSON_SCHEMA: Readonly<Record<string, unknown>> = Object.freeze(
```
caiotheodoro: nit: `Object.freeze` is shallow — the nested `properties`/`required` of this shared singleton (sent as `body.format` on every request) stay mutable, so the freeze intent isn't achieved. Deep-freeze the tree or hand each request a `structuredClone`.

caiotheodoro (commented):
One bug and three risks inline; the mechanism itself is sound (zod-derived schema, flag reachable from JSON configs, dependency in the right package). PR-wide: this claims the same `body.format` slot as #51 with a competing flag and a competing schema (full ActionIntent incl. engine-stamped fields vs #51's model-only packet) — decide the mechanism once before merging either; the strongest merge of the two is a zod-derived packet schema. Also #67 is stacked on this branch (contains all three commits plus four), so #65 merges first or #67 sweeps this in unreviewed.

packages/shared/src/scenario/scenario.types.ts
yoarajota: dropped, wasnt supposed to land on this branch, will go out in its own PR

packages/server/src/llm/ollama-provider.ts
yoarajota: droped the constrainedDecoding branch entirely and converged onto #51s mechanism instead, so this line and the >=0.5 comment dont exist anymore on this branch. the format-drop risk on qwen3+think:false still applies to mains always-on path but thats preexisting from #51, not something this PR adds — worth its own issue

packages/eval/src/run/scenario-runner.ts
yoarajota: removed, this branch doesnt touch scenario-runner.ts anymore

packages/shared/src/intent/intent-json-schema.ts
yoarajota: nice catch, dropped the full-ActionIntent schema and converged onto #51s ModelIntentPacketSchema instead, which already excludes id/actorId from whats requested from the model

packages/shared/src/__tests__/intent-json-schema.test.ts
yoarajota: moved the check into intent-packet-schema.test.ts, pins full enum equality now instead of containment

packages/shared/src/intent/intent-json-schema.ts
yoarajota: dropped that file. mains ModelIntentPacketJsonSchema was never runtime-frozen either, just `as const` at the type level, so no regression here either way

yoarajota:
converged onto #51s mechanism instead of running two competing ones. dropped the full-ActionIntent schema + constrainedDecoding flag + the unrelated scenario files, kept only the zod-to-json-schema derivation applied to #51s already-shipped ModelIntentPacketSchema — thats the 'strongest merge' you mentioned, packet scoping from #51 + zod-derived from here. #67 stacking is on you to sort out after this merges, not something I can fix from in here

yoarajota:
@copilot resolve the merge conflicts in this pull request

matheusht:
@caiotheodoro ping — all five findings landed: constrainedDecoding wired through parseLlmConfig + eval local mode, schema frozen, overstated claim softened, broken test import fixed, zod floor bumped to the peer minimum.

PR #67, opened 2026-08-22
feat(agent): derive the SECTION 8 output contract from ActionIntentSchema

matheusht:
## What

Removes the prompt layer's second source of truth: the SECTION 8 JSON contract is now **derived from `ActionIntentSchema`** instead of hand-written prose that had already drifted from it.

- New `output-contract.ts`: field list, types, and enum choices render mechanically from the derived JSON schema; curated semantic notes are keyed by schema property name so unknown fields get a generic line rather than silence.
- Engine-stamped structure leaves the model's job per the repo's "engine owns structure, model owns language" rule: `id`/`actorId` are never requested (the parser already stamps them), and the old literal delegations ("copy from the inputs", `preferredDelay: 0`) are gone. `fallbackIfBlocked` remains advertised — it is a genuine model decision consumed by the resolver.
- **Drift-proof CI**: a guard test pins set-equality between schema properties and rendered contract fields in both directions (missing or extra field = red), asserts every enum choice renders exhaustively, and verifies no rendered field name exists outside the schema.

Deferred to their own atomic changes (noted in the issue thread): full semantic decision-packet shrink, bounded zod-feedback repair loop, transcript-renderer dedup across judge surfaces, token-estimate hardening.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS 816 passed (+4 drift-guard tests; existing prompt suites unchanged and green)

### DIFF
--- a/packages/server/src/agent/__tests__/output-contract.test.ts
+++ b/packages/server/src/agent/__tests__/output-contract.test.ts
@@ -1,8 +1,6 @@
 import { describe, it, expect } from "vitest";
-import { renderIntentOutputContract } from "../output-contract.js";
-import { ACTION_INTENT_JSON_SCHEMA } from "@perfectman/shared";
-
 import { renderIntentOutputContract, FIELD_NOTES } from "../output-contract.js";
+import { ACTION_INTENT_JSON_SCHEMA } from "@perfectman/shared";
 
 const properties = (ACTION_INTENT_JSON_SCHEMA.properties ?? {}) as Record<
   string,
--- a/packages/server/src/agent/__tests__/output-contract.test.ts
+++ b/packages/server/src/agent/__tests__/output-contract.test.ts
@@ -1,0 +1,48 @@
+import { describe, it, expect } from "vitest";
+import { renderIntentOutputContract } from "../output-contract.js";
+import { ACTION_INTENT_JSON_SCHEMA } from "@perfectman/shared";
+
+const properties = (ACTION_INTENT_JSON_SCHEMA.properties ?? {}) as Record<
+  string,
+  Record<string, unknown>
+>;
+
+describe("renderIntentOutputContract (#49 drift guard)", () => {
+  const contract = renderIntentOutputContract("");
+
+  it("renders every non-engine-stamped schema field exactly once", () => {
+    for (const [name, prop] of Object.entries(properties)) {
+      if (name === "id" || name === "actorId") continue;
+      const needle = `"${name}":`;
+      const count = contract.split(needle).length - 1;
+      expect(count, `${needle} appears once`).toBeGreaterThanOrEqual(1);
+      expect(prop).toBeDefined();
+    }
+  });
+
+  it("never requests engine-stamped fields or hardcodes their values", () => {
+    expect(contract).not.toContain('"id":');
+    expect(contract).not.toContain('"actorId":');
+    // The old hardcoded prose asked the model to invent ids and emit
+    // 'preferredDelay: 0' — those literal delegations must stay gone.
+    expect(contract).not.toContain('"preferredDelay": 0');
+    expect(contract).not.toContain("copy from the inputs");
+  });
+
+  it("renders enum choices exhaustively", () => {
+    const intentEnum = properties.intentType?.enum as string[];
+    expect(intentEnum).toBeDefined();
+    for (const value of intentEnum) {
+      expect(contract, `enum value ${value}`).toContain(`"${value}"`);
+    }
+  });
+
+  it("documents every field either via note or type placeholder", () => {
+    // Every rendered line must correspond to a real schema property —
+    // catches typos in curated notes drifting away from the schema.
+    const rendered = [...contract.matchAll(/"([a-zA-Z]+)":/g)].map(m => m[1]!);
+    for (const name of rendered) {
+      expect(properties[name], `"${name}" is a schema property`).toBeDefined();
+    }
+  });
+});
--- a/packages/server/src/agent/__tests__/output-contract.test.ts
+++ b/packages/server/src/agent/__tests__/output-contract.test.ts
@@ -24,9 +24,8 @@ describe("renderIntentOutputContract (#49 drift guard)", () => {
   it("has no dead curated notes (reverse drift guard)", () => {
     // A note whose schema property was renamed/removed must fail here —
     // otherwise the notes map silently stops rendering.
-    for (const key of Object.keys(FIELD_NOTES)) {
-      expect(properties[key], `FIELD_NOTES.${key} is a schema property`).toBeDefined();
-    }
+    const deadNotes = Object.keys(FIELD_NOTES).filter(k => !(k in properties));
+    expect(deadNotes).toEqual([]);
   });
 
   it("never requests engine-stamped fields or hardcodes their values", () => {
@@ -41,7 +40,7 @@ describe("renderIntentOutputContract (#49 drift guard)", () => {
   it("renders enum choices exhaustively", () => {
     for (const enumField of ["intentType", "channelType", "fallbackIfBlocked"]) {
       const values = properties[enumField]?.enum as string[] | undefined;
-      expect(values, `${enumField} is an enum in the schema`).toBeDefined();
+      expect(Array.isArray(values), `${enumField} is an enum in the schema`).toBe(true);
       for (const value of values!) {
         expect(contract, `${enumField} value ${value}`).toContain(`"${value}"`);
       }
@@ -52,8 +51,7 @@ describe("renderIntentOutputContract (#49 drift guard)", () => {
     // Every rendered line must correspond to a real schema property —
     // catches typos in curated notes drifting away from the schema.
     const rendered = [...contract.matchAll(/"([a-zA-Z]+)":/g)].map(m => m[1]!);
-    for (const name of rendered) {
-      expect(properties[name], `"${name}" is a schema property`).toBeDefined();
-    }
+    const foreign = rendered.filter(name => !(name in properties));
+    expect(foreign).toEqual([]);
   });
 });
--- a/packages/server/src/agent/__tests__/output-contract.test.ts
+++ b/packages/server/src/agent/__tests__/output-contract.test.ts
@@ -2,6 +2,8 @@ import { describe, it, expect } from "vitest";
 import { renderIntentOutputContract } from "../output-contract.js";
 import { ACTION_INTENT_JSON_SCHEMA } from "@perfectman/shared";
 
+import { renderIntentOutputContract, FIELD_NOTES } from "../output-contract.js";
+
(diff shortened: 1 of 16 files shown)


packages/server/src/agent/__tests__/output-contract.test.ts
```diff
@@ -0,0 +1,57 @@
+import { describe, it, expect } from "vitest";
+import { renderIntentOutputContract } from "../output-contract.js";
+import { ACTION_INTENT_JSON_SCHEMA } from "@perfectman/shared";
+
+import { renderIntentOutputContract, FIELD_NOTES } from "../output-contract.js";
```
caiotheodoro, line 2: bug: `renderIntentOutputContract` is imported at line 2 and again here — invalid TS/ESM that only runs because tests are excluded from typecheck and vitest 2.1.9's esbuild transform tolerates it. On vitest ≥3 (oxc) the whole file fails at transform with zero tests run, silently disabling the drift guard this PR exists to add. Merge the two into one import.

packages/server/src/agent/output-contract.ts
```diff
@@ -0,0 +1,104 @@
...
+    "ONLY for send_message/reply_to_message/leave_channel/invite_agent — the existing channel ID you're acting in. OMIT entirely for create_channel (use channelName/channelType instead).",
+  personTargets:
+    "OMIT this field unless the action targets a specific person directly.",
+  visibleContent:
+    "Optional text message content (required for 'send_message' or 'reply_to_message')",
+  privateMotiveSummary:
+    "A highly specific natural-language thought explaining your real, raw, hidden motive behind this action. Never leave empty.",
+  emotionDrivers: ["warmth", "jealousy", "irritation"],
+  motivationDrivers: ["affinity", "gossip", "exclusion"],
+  fallbackIfBlocked:
```
caiotheodoro, line 36: risk: this note promises "if your primary action is denied, the low-risk action to take instead" and the enum advertises all 11 intent types — but nothing consults `fallbackIfBlocked` when an action is blocked; only intent-parser's parse-FAILURE path reads it, honoring exactly `no_op` and `delay_response`. 9 of 11 advertised choices are dead tokens. Reword to: safety fallback if your reply can't be parsed — only 'no_op' or 'delay_response' are honored. (If #64 merges, revisit.)

packages/server/src/agent/output-contract.ts
```diff
@@ -0,0 +1,104 @@
...
+ * structure, the model owns language (see intent-parser's structural
+ * repairs, which already stamp them). Everything the model IS asked for is
+ * either semantic natural language or a choice among runtime options.
+ */
+
+/** Fields the engine stamps after parsing — never requested from the model. */
+const ENGINE_STAMPED = new Set(["id", "actorId"]);
+
+/** Curated per-field guidance, keyed by schema property name. String arrays are example lists. */
+export const FIELD_NOTES: Record<string, string | string[]> = {
```
caiotheodoro, line 24: risk: `preferredDelay` has no curated note so it renders as a bare `<number>` — no units, no semantics — where the old contract's hardcoded `0` neutralized the resolver behavior that delays any intent with `preferredDelay > 0` by `ceil(ms / pulseIntervalMs)` pulses. A small model emitting 3 or 1500 on an ordinary send_message silently defers its own message. Add a FIELD_NOTES entry ("OMIT unless you deliberately want to wait — milliseconds") or exclude the field like id/actorId.

packages/server/src/agent/output-contract.ts
```diff
@@ -0,0 +1,104 @@
...
+    "Optional text message content (required for 'send_message' or 'reply_to_message')",
+  privateMotiveSummary:
+    "A highly specific natural-language thought explaining your real, raw, hidden motive behind this action. Never leave empty.",
+  emotionDrivers: ["warmth", "jealousy", "irritation"],
+  motivationDrivers: ["affinity", "gossip", "exclusion"],
+  fallbackIfBlocked:
+    "if your primary action is denied, the low-risk action to take instead (optional)",
+  channelName: "ONLY for create_channel — a short name for the new channel.",
+  channelType: "ONLY for create_channel — usually private_channel.",
+  invitedAgentIds:
```
caiotheodoro, line 40: risk: string notes render in value position and erase array shape — `"invitedAgentIds": "ONLY for create_channel..."` shows an array field as a quoted string, and unlike personTargets/memoryWrites, invitedAgentIds is NOT array-coerced by the parser's repair (only null-deleted), so a model copying the rendered shape gets a zod reject and the intent collapses to fallback. In renderPlaceholder, wrap string notes for array-typed props as `["note"]`.

packages/server/src/agent/output-contract.ts
```diff
@@ -0,0 +1,104 @@
...
+  return `"${fieldName}": ...`;
+}
+
+export function renderIntentOutputContract(ownUtterancesWarning: string): string {
+  const properties = (ACTION_INTENT_JSON_SCHEMA.properties ?? {}) as Record<
+    string,
+    Record<string, unknown>
+  >;
+
+  const requested = Object.keys(properties)
```
caiotheodoro, line 78: nit: the renderer never reads the schema's `required` array — required-ness comes from curated prose, so a future required field without a note renders unmarked. #51's contract prints (required)/(optional) mechanically; same trick would complete the single-source-of-truth claim.

packages/server/src/agent/output-contract.ts
```diff
@@ -0,0 +1,104 @@
...
+    "A highly specific natural-language thought explaining your real, raw, hidden motive behind this action. Never leave empty.",
+  emotionDrivers: ["warmth", "jealousy", "irritation"],
+  motivationDrivers: ["affinity", "gossip", "exclusion"],
+  fallbackIfBlocked:
+    "if your primary action is denied, the low-risk action to take instead (optional)",
+  channelName: "ONLY for create_channel — a short name for the new channel.",
+  channelType: "ONLY for create_channel — usually private_channel.",
+  invitedAgentIds:
+    "ONLY for create_channel — agentIds to invite into the new channel.",
+  memoryWrites:
```
caiotheodoro, line 42: nit: the memoryWrites note hand-duplicates MemoryWriteProposalSchema's nested fields and type enum as prose — the second source of truth this PR eliminates, one level down — and the drift guard only checks top level, so a nested rename drifts silently. Derive from `items.properties` or note the gap in the header.

caiotheodoro (commented):
One bug and three risks inline on the delta (reviewed as the diff over #65, since this branch is stacked on it — merge #65 first or this sweeps it in unreviewed). PR-wide: all three delta files sit in the layer #51 deletes — #51's modelIntentPacketFieldContract does this same derivation from the packet schema, which engine-stamps preferredDelay and fallbackIfBlocked. If #51 lands first this delta's surviving value is roughly zero as code; the salvageable pieces are the curated FIELD_NOTES prose and the reverse-drift-guard idea, both portable into #51's contract. Worth deciding the sequencing now rather than after both invest further. Also the PR body's justification is inaccurate: the resolver never reads fallbackIfBlocked — only intent-parser's parse-failure path does, honoring just no_op/delay_response.

matheusht:
@caiotheodoro ping — final reviewer verdict landed: APPROVE. Both hygiene items from that pass are fixed (duplicate import merged, memoryWrites rendering note corrected). All prior rounds' findings hold at HEAD: balanced rendered contract, exhaustive enums, bidirectional drift guard.

yoarajota:
soo, main already shipped this one in `19f7500` — the packet contract is derived from zod, `modelIntentPacketFieldContract()` emits the `(required)`/`(optional)` markers, and `composeIntentPacket()` stamps `preferredDelay`/`fallbackIfBlocked` engine-side, so the `SECTION 8` string and 3 of the 6 review comments are gone already... the branch also carries two things main replaced: the `constrainedDecoding` config (off by default, ollama-native `format` only) vs main's `responseFormatJsonSchema` (on by default, openai-compatible, 400/422 fallback), and a stray `packages/shared/src/scenario/` thats a duplicate of the existing scenarios lib and never got exported. whats genuinely not on main is the `FIELD_NOTES` prose in `server/src/agent/output-contract.ts` — `channelTarget` "only for send_message/reply/leave/invite, omit for create_channel", `emoji` "only for react", `replyToEventId`, `targetEventId`, `invitedAgentIds` — main's shape-only renderer says nothing about per-type applicability, and that guidance is the part worth keeping. three ways forward: (1) close this and open a small follow-up layering the notes over `modelIntentPacketFieldContract()`; (2) reduce this PR in place to just that — drop the two llm commits, `shared/src/scenario/`, `intent-json-schema.ts` and the scenario-runner hunk, then rewrite output-contract as a notes layer on main's contract; or (3) keep the branch whole and revert main's approach — i'd argue against that one, main's is strictly more capable and already has consumers. my rec is (2), keeps your authorship of the prose. not closing or touching the branch until you pick, no?

yoarajota (changes requested):
.

packages/shared/src/intent/intent-json-schema.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * JSON Schema derived from ActionIntentSchema — the single source of truth
+ * (see #49's drift concern: hand-written contract copies are how prompt and
+ * validation drifted apart historically).
+ *
+ * Consumed by OllamaProvider as a structured-output `format` target, so the
+ * model can only emit tokens that fit the intent shape at decoding time.
+ */
+export const ACTION_INTENT_JSON_SCHEMA: Readonly<Record<string, unknown>> = Object.freeze(
+  zodToJsonSchema(ActionIntentSchema, {
```
caiotheodoro, line 13: risk: this schema is the full `ActionIntentSchema` (id, actorId, preferredDelay, fallbackIfBlocked included). Constrained decoding would force the model to emit the fields the prompt says the engine stamps. Main's `ModelIntentPacketJsonSchema` is the model-only packet; derive from that, or drop the llm commits with the rest of the branch.

caiotheodoro (changes requested):
Close it. Main already derives the packet contract from zod (`modelIntentPacketFieldContract`), engine-stamps `preferredDelay`/`fallbackIfBlocked`, and #90 sends `ModelIntentPacketJsonSchema` as Ollama `format` with a 400/422 json fallback. One inline. Salvage is the per-type FIELD_NOTES (channelTarget / invitedAgentIds applicability) layered onto `modelIntentPacketFieldContract()` — not this branch.

caiotheodoro:
Closing: main already derives the packet contract from zod (`modelIntentPacketFieldContract`), engine-stamps `preferredDelay`/`fallbackIfBlocked`, and #90 sends `ModelIntentPacketJsonSchema` as Ollama `format`. Don't rebase this. Salvage is the per-type FIELD_NOTES (channelTarget / invitedAgentIds) as a notes layer on `modelIntentPacketFieldContract()`, not this branch.

Issue #48, opened 2026-08-20
Memory retrieval is flat recency+relevance — SOTA uses hierarchical/graph-based retrieval for long-horizon coherence


caiotheodoro:
`build-perception-packet.ts`'s memory retrieval (`relevantMemories`) is a flat sort by involved-people-relevance then recency, capped at 8 memories (`MAX_MEMORIES`) — no clustering, summarization, or graph structure over the memory store. Recent memory-augmented-agent research (RAPTOR's recursive multi-granularity summaries, H-MEM's hierarchical memory routing, SGMem's sentence-graph memory) shows flat vector/recency retrieval degrades as the memory store grows — "once the most relevant evidence has been recalled, adding more semantically similar memories often brings diminishing retrieval precision, inflating context cost" without necessarily improving coherence.

Distinct from #34 (which covers the *write*/consolidation side — the still-stubbed `background_reflection` pass): this is about the *retrieval* side, which is already wired and live but using the simplest possible strategy. Most relevant for long-running scenarios (the 90-pulse novela scenario) where an agent's memory store could grow large enough for flat recency-sorting to start missing genuinely important older memories in favor of merely-recent ones.

**Where to look**: `packages/engine/src/perception/build-perception-packet.ts` (`MAX_MEMORIES`, the current sort), `packages/shared/src/memory/memory.types.ts` (the `Memory` type — already has `type`/`subjectAgentIds`/`confidence` fields that a smarter retrieval strategy could use beyond plain recency).

matheusht:
## Plan

Replace the flat (relevanceFlag, recency) sort with a single transparent salience score per memory — no clustering/graph machinery yet (the store is capped far below where RAPTOR/H-MEM-style structure pays off; this is the cheapest step that stops recency from evicting important older memories):

```
salience = 2.0·relevant(subject ∩ involvedPeople)
         + typeWeight   (pending_intention 1.0 > relationship 0.8 > self 0.5 >
                         social_theory/emotional_residue 0.4 > episodic 0.3)
         + confidence   (0..1, authored at write time)
         + 0.25·unresolved
         + 0.5·recency  ((newestAge - age)/newestAge, monotonic, bounded)
```

- Pure function `scoreMemorySalience()` in engine (stays I/O-free); selection = sort desc + existing MAX_MEMORIES=8 cap.
[...]

PR #68, opened 2026-08-22
feat(engine): salience-scored memory retrieval in perception packets

matheusht:
## What

Upgrades perception-packet memory retrieval from a flat recency sort to salience scoring:

- New pure module `memory-salience.ts`: each memory gets one transparent score from signals the `Memory` type already carries — subject relevance against the current pulse's involved people (dominant weight), a memory-type tier (`pending_intention` > `relationship` > `self` > `social_theory`/`emotional_residue` > `episodic`), authored confidence, an open-loop bonus for unresolved memories, and recency demoted to a bounded linear decay instead of the deciding key.
- `buildPerceptionPacket` selects by score (same 8-memory cap). An old unresolved high-confidence intention about the person currently addressing you now outranks a fresh low-confidence aside — the exact failure mode of recency-first sorting in long scenes.
- Deterministic and I/O-free: stable createdAt tiebreak, fallback weight for unknown types, no new dependencies.
- Ten unit tests: relevance dominance, open-loop precedence, type-tier ordering, confidence effect, recency bound proof (max delta = 0.5), empty store, cap behavior, salience ranking, tie determinism.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS 832 passed (+10)
- Mock bench slice over memory-centric scenarios holds signal pass at 100%.


packages/engine/src/perception/memory-salience.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export function scoreMemorySalience(
+  memory: Memory,
+  involvedPeople: ReadonlySet<string>,
+  newestCreatedAt: number,
+): number {
+  const relevant = memory.subjectAgentIds.some(id => involvedPeople.has(id)) ? 1 : 0;
+  const typeWeight = MEMORY_TYPE_WEIGHTS[memory.type] ?? 0.3;
+
+  const ageSpan = Math.max(1, newestCreatedAt - Math.min(memory.createdAt, newestCreatedAt));
```
caiotheodoro: bug: `ageSpan` is per-memory (`max(1, newest − this.createdAt)`), so recency = `1 − age/age` = 0 for anything ≥1ms older than the newest memory. The documented linear decay is actually a binary newest-or-nothing: a memory from 1 second ago and one from 3 days ago score identically. Normalize over the store's span instead — compute `oldestCreatedAt` alongside `newestCreatedAt` in `selectRelevantMemories`, then `recency = 1 − min(1, (newest − createdAt) / max(1, newest − oldest))`.

packages/engine/src/perception/__tests__/memory-salience.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    const low = scoreMemorySalience(memory({ confidence: 0.1 }), involved, 100);
+    expect(high).toBeGreaterThan(low);
+  });
+
+  it("keeps recency as a bounded term, not a dominating one", () => {
+    const involved = new Set<string>();
+    // Newest possible vs oldest possible differs by RECENCY_WEIGHT only.
+    const newest = scoreMemorySalience(memory({ createdAt: 100 }), involved, 100);
+    const oldest = scoreMemorySalience(memory({ createdAt: 0 }), involved, 100);
+    expect(newest - oldest).toBeCloseTo(0.5);
```
caiotheodoro, line 83: risk: this test pins only the two extremes (age 0, age max), exactly where the broken binary decay coincides with true linear decay — it's the test that should have caught the bug above. Add a midpoint: with newest=100, a `createdAt: 50` memory must score strictly between the 0 and 100 ones (today it ties with 0).

packages/engine/src/perception/memory-salience.ts
```diff
@@ -0,0 +1,64 @@
...
+export const MEMORY_TYPE_WEIGHTS: Record<Memory["type"], number> = {
+  pending_intention: 1.0,
+  relationship: 0.8,
+  self: 0.5,
+  social_theory: 0.4,
+  emotional_residue: 0.4,
+  episodic: 0.3,
+};
+
+const RELEVANCE_WEIGHT = 2.0;
```
caiotheodoro: risk: the PR body calls subject relevance the dominant weight, but 2.0 doesn't dominate — the max opposing delta is 2.45 (0.7 type + 1.0 confidence + 0.25 unresolved + 0.5 recency), so an irrelevant fresh high-confidence pending_intention outranks a relevant low-confidence episodic. That inverts the old sort's relevance-first guarantee. Either make the claim true or document the blend as intended:
```suggestion
const RELEVANCE_WEIGHT = 2.5; // > max opposing delta (0.7 + 1.0 + 0.25 + 0.5) so relevance strictly dominates
```

packages/engine/src/perception/build-perception-packet.ts
```diff
@@ -85,17 +86,16 @@ export function buildPerceptionPacket(
...
-  const involvedSet = involvedPeople;
-  const sortedMemories = [...agent.memories].sort((a, b) => {
-    // Prioritize memories about involved people
-    const aRelevant = a.subjectAgentIds.some(id => involvedSet.has(id)) ? 1 : 0;
-    const bRelevant = b.subjectAgentIds.some(id => involvedSet.has(id)) ? 1 : 0;
-    if (aRelevant !== bRelevant) return bRelevant - aRelevant;
-    return b.createdAt - a.createdAt; // newer first
-  });
-  const relevantMemories = sortedMemories.slice(0, MAX_MEMORIES);
+  const relevantMemories = selectRelevantMemories(
```
caiotheodoro, line 94: risk: the claimed failure-mode fix is pinned only at the pure-function level — no test exercises it through `buildPerceptionPacket`, and the existing golden assertion (>0 memories, ≤8 cap) also passed under the old sort. One packet-level test with >8 memories where a salient old memory must survive the cap would make integration regressions catchable.

packages/engine/src/perception/memory-salience.ts
```diff
@@ -0,0 +1,64 @@
...
+  memory: Memory,
+  involvedPeople: ReadonlySet<string>,
+  newestCreatedAt: number,
+): number {
+  const relevant = memory.subjectAgentIds.some(id => involvedPeople.has(id)) ? 1 : 0;
+  const typeWeight = MEMORY_TYPE_WEIGHTS[memory.type] ?? 0.3;
+
+  const ageSpan = Math.max(1, newestCreatedAt - Math.min(memory.createdAt, newestCreatedAt));
+  // Memories created at `newestCreatedAt` get 1; older ones decay linearly to 0.
+  const ageOfThis = Math.max(0, newestCreatedAt - memory.createdAt);
```
caiotheodoro, line 39: question: scoring ignores `lastReinforcedAt`, which `Memory` carries and the biased-memory fixture deliberately sets — a frequently-reinforced old memory ages as if never touched, which cuts against #48's long-horizon goal. Was `max(createdAt, lastReinforcedAt)` as the recency anchor considered?

packages/engine/src/perception/__tests__/memory-salience.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    simulationId: "sim",
+    type: "episodic",
+    subjectAgentIds: [],
+    sourceEventIds: [],
+    summary: "something happened",
+    emotionalTone: "neutral",
+    confidence: 0.5,
+    unresolved: false,
+    createdAt: 100,
+    lastReinforcedAt: null,
```
caiotheodoro: nit: `lastReinforcedAt: null` contradicts the shared `Memory` type (`number`); it only compiles because engine's tsconfig excludes tests from typecheck.
```suggestion
    lastReinforcedAt: 100,
```

packages/engine/src/perception/__tests__/memory-salience.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        unresolved: true,
+        createdAt: 40,
+      }),
+      memory({ id: "noise-3", createdAt: 103 }),
+    ];
+    const selected = selectRelevantMemories(store, new Set(["bruno"]), 2);
+    expect(selected[0]!.id).toBe("open-intent");
+  });
+
+  it("breaks exact ties by recency deterministically", () => {
```
caiotheodoro, line 133: nit: this test never reaches the `createdAt` tiebreak it names — the recency term already separates the scores. Equal scores with unequal `createdAt` (e.g. offset confidence to cancel recency) is what hits the branch.

caiotheodoro (commented):
One bug inline (recency decay is binary, not linear) plus the test gap that let it pass, and a weight-dominance question that inverts an old guarantee. PR-wide: only non-regression evidence is claimed and none is committed — docs/eval/evidence/ exists for exactly this, and #48's motivating case (late novela pulses where the store outgrows the 8-cap) is the comparison that would show salience actually beats recency. Also: this is a weighted linear score, not the hierarchical retrieval #48 asks for — say `Refs #48 (partial)` rather than letting the merge retire the issue.

Issue #35, opened 2026-08-20
Repetition guard's Jaccard threshold (0.7) and single-retry policy are unvalidated


caiotheodoro:
The structural repetition guard added in PR #22 (`packages/server/src/agent/repetition-guard.ts`) uses a Jaccard word-overlap similarity threshold of 0.7 and allows exactly one retry before falling back to `no_op`. Both numbers were chosen pragmatically to ship a working fix, not tuned against real data.

Worth an experiment sweep: try different thresholds (looser/stricter) and retry counts (0, 1, 2+) against the same benchmark scenarios, and compare narrative_cohesion scores + real-content-vs-blocked-turn ratios. A threshold too loose won't catch near-repeats; too strict will false-positive on legitimately similar-but-distinct in-character responses (e.g. a persona who says "kkkk" a lot).

**Where to look**: `packages/server/src/agent/repetition-guard.ts` (`REPETITION_SIMILARITY_THRESHOLD`, `isNearRepeat`), `packages/server/src/agent/agent-runtime.ts`'s retry block, `packages/server/src/agent/__tests__/repetition-guard.test.ts` for the existing test cases to extend.

matheusht:
## Plan

Make the guard's two pragmatic constants measurable instead of sacred:

1. `AgentRuntime` gains an optional `RepetitionPolicy` ({threshold?, maxRetries?}) — defaults stay exactly today's behavior (0.7 / one retry). The retry block becomes a bounded loop; `maxRetries: 0` blocks a detected repeat immediately.
2. Eval `RunnerOpts` threads the policy so in-process sweeps can vary it per run.
3. New offline sweep CLI (`pnpm --filter @perfectman/eval sweep:repetition`): grid thresholds × retry counts over chatty/memory-heavy scenarios; per-cell metrics from real artifacts — guard-block rate, content-repetition probe mean, rule-judge narrative_cohesion, LLM-call totals. Writes a JSON matrix; mock-mode dry-run evidence committed under docs/eval/evidence/. Real-model sweeps remain documented maintainer follow-ups.
4. Unit tests: two-retry recovery (3 calls), zero-retry immediate block (1 call), loose vs strict threshold behavior at the boundary.

Closes #35 once shipped.

yoarajota:
[...]

PR #69, opened 2026-08-22
feat(agent): repetition-guard policy knobs + offline threshold/retry sweep

matheusht:
## What

Turns the repetition guard's two pragmatic constants (Jaccard threshold 0.7, one retry) into measured, configurable policy:

- `AgentRuntime` accepts an optional `RepetitionPolicy` — `{ threshold, maxRetries }` with defaults reproducing shipped behavior byte-for-byte. The retry block is now a bounded loop; a zero retry budget blocks a detected repeat on the spot instead of silently committing it.
- Eval's `ScenarioRunner` threads the policy through to the runtime, enabling in-process sweeps.
- New offline sweep CLI (`pnpm --filter @perfectman/eval sweep:repetition`): runs the thresholds × retries grid over chatty/memory-heavy scenarios and records per-cell guard-block counts, content-repetition probe means, rule-judge `narrative_cohesion`, and LLM-call totals to a JSON matrix.
- Committed mock-mode baseline is honestly flat and says so: the deterministic mock regenerates identical canned intents regardless of correction notes, and its repeats sit at ~1.0 Jaccard — every cell catches exactly the same set. Discriminating between cells needs real-model runs; that protocol is the documented maintainer follow-up.
- Four new runtime tests prove each knob: two-retry recovery on the third attempt (3 provider calls), zero-budget immediate block (1 call), strict-threshold pass-through of a loose near-match, lowered-threshold blocking.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS 826 passed (+4 knob tests)
- Sweep CLI executed end-to-end offline; evidence matrix committed under docs/eval/evidence/.


packages/server/src/agent/__tests__/agent-runtime.test.ts
```diff
@@ -271,5 +271,86 @@ describe("AgentRuntime Orchestration", () => {
...
+
+      const output = await runtime.generateIntent(repeatingInput, context);
+
+      expect(fetchMock).toHaveBeenCalledTimes(1);
+      expect(output.fallbackApplied).toBe(true);
+      expect(output.intent.intentType).toBe("no_op");
+      expect(output.intent.privateMotiveSummary).toContain("Repetition guard");
+    });
+
+    it("honors a stricter threshold: a loose near-match passes when threshold is raised", async () => {
```
caiotheodoro, line 315: bug: this raised-threshold test is vacuous — the fixture pair sits at Jaccard 4/6 ≈ 0.667, already below the 0.7 default, so it passes even if the threshold knob is never plumbed. Use a candidate strictly between default and raised threshold, e.g. `'kkkkk não acredito nesse take mesmo'` (5/6 ≈ 0.833): blocks at 0.7, passes at 0.99. The lowered-threshold sibling at line 338 is fine.

examples/simulations/README.md
```diff
@@ -5,6 +5,7 @@ Copy one of these files to `config/index.json` and edit locally.
 - `mock.inline-personas.example.json` — in-memory simulation with personas embedded directly in the config.
 - `mock.persona-file.example.json` — in-memory simulation loading a compiled persona file from `config/personas/`.
 - `freellmapi.example.json` — example for a FreeLLMAPI-compatible provider.
+- `deepseek.example.json` — example for DeepSeek's OpenAI-compatible API; requires `DEEPSEEK_API_KEY`.
```
caiotheodoro: bug: this documents `deepseek.example.json`, which exists neither on this branch nor on main — stray hunk from another branch. Drop the line or add the file.

packages/eval/src/cli/sweep-repetition.ts
```diff
@@ -0,0 +1,130 @@
...
+        const artifact = await ScenarioRunner.run(scenario!, {
+          llmMode: "mock",
+          repetition: { threshold, maxRetries },
+        });
+        cell.runs++;
+        cell.guardBlocks += countGuardBlocks(artifact.events);
+        for (const calls of artifact.llmCalls.values()) cell.llmCalls += calls;
+        cell.providerCalls += artifact.providerCalls ?? 0;
+
+        const repetitionProbe = artifact.probeResults.find(p => p.probe === "content-repetition");
```
caiotheodoro, line 88: risk: `contentRepetitionMean` is measured at the probe's fixed 0.7 (runAllProbes passes no threshold) while the runtime threshold varies per cell — keeping the yardstick fixed is the right call, but nothing in the docstring, report, or evidence says so, and a reader will assume the probe moved with the cell. Record `probeThreshold: REPETITION_SIMILARITY_THRESHOLD` in the output and add one docstring line.

packages/server/src/agent/agent-runtime.ts
```diff
@@ -139,32 +149,50 @@ export class AgentRuntime {
...
-          intent = retryParse.intent;
-          fallbackApplied = retryParse.fallbackApplied;
-        } else {
+      const maxRetries = Math.max(0, Math.floor(this.repetitionPolicy?.maxRetries ?? 1));
+      if (maxRetries <= 0) {
+        // No retry budget: the detected repeat blocks right away.
+        repetitionBlocked = true;
+      }
+      let lastAttemptContent = intent.visibleContent;
+      for (let attempt = 0; attempt < maxRetries; attempt++) {
```
caiotheodoro: risk: retries bypass the budget — `canCall` runs once before the first call, the loop issues up to maxRetries wire calls with no pre-check, and one merged usage record makes accounting see 1 call for 1+N. At maxRetries=1 that was the shipped status quo, but this knob scales the unmetered overage linearly (the sweep itself shows +61 wire calls per retry step). Re-check `canCall` inside the loop, or at minimum say in the RepetitionPolicy JSDoc that retries aren't budget-checked.

docs/eval/evidence/repetition-sweep-mock.json
```diff
@@ -0,0 +1,102 @@
+{
+  "version": "repetition-sweep-v1",
+  "mode": "mock",
```
caiotheodoro, line 3: risk: this matrix is flat across the threshold axis by construction (mock repeats sit at ~1.0 Jaccard), so it validates nothing about 0.7 and can't surface #35's catchphrase false-positive concern — the PR body says so, the artifact doesn't, and future readers of docs/eval/evidence/ will cite it as the study that justified the defaults. Embed a `limitations` field saying threshold cells are indistinguishable by construction and only providerCalls varies.

packages/server/src/agent/agent-runtime.ts
```diff
@@ -173,7 +201,7 @@ export class AgentRuntime {
       intent = IntentParser.createFallback(
         agentId,
         "no_op",
-        "Repetition guard: near-duplicate of a message you already sent, even after a retry — blocked structurally.",
+        `${REPETITION_GUARD_MARKER}: near-duplicate of a message you already sent, even after a retry — blocked structurally.`,
```
caiotheodoro: nit: with maxRetries=0 the block motive still says "even after a retry" when none was attempted — and this string is now load-bearing (the sweep greps it). Word it from the actual attempt count.

packages/server/src/agent/agent-runtime.ts
```diff
@@ -139,32 +149,50 @@ export class AgentRuntime {
...
+      const maxRetries = Math.max(0, Math.floor(this.repetitionPolicy?.maxRetries ?? 1));
+      if (maxRetries <= 0) {
+        // No retry budget: the detected repeat blocks right away.
+        repetitionBlocked = true;
+      }
+      let lastAttemptContent = intent.visibleContent;
+      for (let attempt = 0; attempt < maxRetries; attempt++) {
+        const retryPrompt = {
+          ...prompt,
+          system: `${prompt.system}\n\nIMPORTANT: your last attempt this turn ("${lastAttemptContent}") was too close to something you already said. Say something genuinely different — a new angle, a reaction to someone else, a topic change — or choose "no_op" if you truly have nothing new to add.`,
```
caiotheodoro: nit: the retry prompt never escalates — attempt 2 gets the identical instruction with only the quoted attempt swapped. A model that ignored it once will likely ignore its twin; include the attempt count or harden later attempts. Fine to defer, mock data can't measure it either way.

caiotheodoro (commented):
Two bugs and three risks inline; the mechanics are solid — defaults verifiably reproduce shipped behavior (threshold ?? 0.7 clamped, maxRetries ?? 1, identical retry semantics, no call site changes). PR-wide: the policy is programmatic-only (no JSON config key — worth saying in the body), the probe yardstick correctly stays fixed at 0.7 while the runtime knob moves but that choice is documented nowhere, and as committed this ships the instrument, not the study — the mock matrix can't discriminate thresholds, so #35 should stay open after merge.

matheusht:
@caiotheodoro ping — three review cycles complete, every finding applied (wire-call metric, shared REPETITION_GUARD_MARKER, integer retry budgets, body-routed test stubs). Committed mock evidence honestly flat; live sweeps documented as maintainer follow-up.

Issue #46, opened 2026-08-20
Cross-agent stylistic convergence ("attractor states") — distinct from single-agent repetition


caiotheodoro:
Recent research on multi-turn LLM conversations ("Attractor States Emerge in Multi-Turn LLM Conversations", arXiv:2606.30571) found that open-ended multi-agent dialogue settles into model/persona-specific *attractors* — stable behavior patterns that asymmetrically pull other agents' style toward them, independent of topic. One persona can act as a strong attractor (others start echoing its traits/phrasing) while another is more malleable.

This session observed exactly this pattern empirically, though not by that name: when a worked JSON example was added to the prompt (later reverted, see PR #20's history), **all 5 personas in a scene converged on copying the same phrasing**, not just their own individual repetition. The existing repetition guard (`repetition-guard.ts`, see #35) only checks an agent against *its own* recent utterances — it has no mechanism to detect or correct convergence *across* agents, which is a different failure mode.

Worth investigating: (a) whether this convergence still happens post-fix under normal (non-buggy-example) conditions, at what rate, and which personas act as attractors vs. get pulled; (b) whether a cross-agent similarity check belongs alongside the existing per-agent one.

**Where to look**: `packages/server/src/agent/repetition-guard.ts` (per-agent only, today), the transcript evidence in PR #22's history (the worked-example copying incident), arXiv:2606.30571 for the methodology (self-play vs. mixed-play trajectory comparison in representation space).

matheusht:
## Plan

Make cross-agent convergence a measured number instead of a transcript-reading anecdote:

1. New pure probe `crossAgentEchoRate(events, threshold?)` in the eval probe set: for each content-bearing turn, compute Jaccard similarity (same canonical `isNearRepeat` primitive) against the pooled earlier utterances of every OTHER agent; the turn counts as an echo when any prior cross-agent utterance meets the threshold. Rate = echoed turns / content turns.
2. Seed band `"cross-agent-echo": [0, 0.1]`, calibration-pending like its siblings; wired into `runAllProbes`.
3. Attractor attribution helper `echoSourcesByAgent(events, threshold?)`: per target agent, which agent's prior text it echoed most — the "who pulls whom" question the issue asks, answerable from any bench artifact without new infrastructure.
4. Unit tests: verbatim echo detected across agents, self-repetition NOT counted (that's the existing per-agent probe), distinct content zero, empty transcripts, source attribution correctness.

[...]

PR #70, opened 2026-08-22
feat(eval): cross-agent echo probe with attractor attribution

matheusht:
## What

Makes multi-agent stylistic convergence measurable instead of anecdotal:

- New pure probe `crossAgentEchoRate(events, threshold?)`: share of content-bearing turns that near-repeat (canonical Jaccard primitive, now shared as `similarity()` from the runtime) an **earlier utterance by a different agent**. The per-agent repetition guard and the per-agent content-repetition probe are structurally blind to this failure mode.
- New attribution helper `echoSourcesByAgent(events, threshold?)`: counts keyed `"target<-source"` answering *who pulls whom* directly from any bench artifact — no new infrastructure needed to investigate attractor personas.
- Seed band `cross-agent-echo [0, 0.1]` wired into `runAllProbes`; chronological-order contract documented.
- Detection only by design: whether the runtime should ever *correct* cross-agent echo is left open until measured rates from real runs exist.
- Four new tests (+ one band-wiring assertion): cross-echo detection, self-repeat exclusion (belongs to the other probe), source attribution, empty/single-turn transcripts, band wiring.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (probe suite 19/19)
- Live mock bench on the echo-chamber scenario renders the new row (`cross-agent-echo 0.000 | 100%`) — clean baseline under deterministic generation.


packages/eval/src/probes/index.ts
```diff
@@ -234,6 +234,75 @@ export function contentRepetitionRate(
...
+  );
+  if (contentTurns.length === 0) return 0;
+  const utterancesByAgent = new Map<string, string[]>();
+  let echoed = 0;
+  for (const turn of contentTurns) {
+    const others: string[] = [];
+    for (const [agentId, utterances] of utterancesByAgent) {
+      if (agentId !== turn.agentId) others.push(...utterances);
+    }
+    if (isNearRepeat(turn.content!, others, threshold)) echoed++;
```
caiotheodoro: bug: empty-normalized-word-set turns score similarity 1.0 across agents — `normalizeWords` strips emoji/punctuation/stopwords and `similarity` returns 1 for two empty sets, and the `content?.trim()` filter doesn't exclude them. Ran it: `[A:"😂😂", B:"🔥🔥🔥", C:"kkkk vamos nessa"]` → `crossAgentEchoRate = 1/3` with a fabricated `B<-A` attractor edge; `[A:"que de o a", B:"the and of to"]` → 0.5. In an emoji-heavy sim this inflates a probe banded at [0, 0.1]. Fix at the probe layer: count a turn as content-bearing only if its normalized word set is non-empty (same edge #57's review flagged on this primitive, now cross-agent so any agent pair hits it).

packages/eval/src/probes/index.ts
```diff
@@ -234,6 +234,75 @@ export function contentRepetitionRate(
...
+    e => (e.kind === "post" || e.kind === "reply") && !!e.content?.trim(),
+  );
+  const utterancesByAgent = new Map<string, string[]>();
+  for (const turn of contentTurns) {
+    let bestSource: string | undefined;
+    let bestScore = 0;
+    for (const [agentId, utterances] of utterancesByAgent) {
+      if (agentId === turn.agentId) continue;
+      for (const prior of utterances) {
+        const score = jaccardSimilarity(turn.content!, prior);
```
caiotheodoro: risk: attribution inverts the attractor arrow on mixed repeats. `echoSourcesByAgent` compares a turn only against other agents' priors, never the agent's own — ran `[A:X, B:X, A:X]`: A's turn 3 is a self-repeat of its own turn 1, yet it's attributed as `A<-B`, crediting an inbound edge to the attractor persona. That's precisely the who-pulls-whom analysis #46 asks for, backwards. When a turn is also a self-near-repeat, skip it (leave it to the per-agent probe) or attribute to self and drop the edge; the existing self-repeat test only covers the pure case.

packages/eval/src/test/probes.test.ts
```diff
@@ -158,6 +160,50 @@ describe("probes", () => {
...
+        ev("post", "A", 1, "fala sério que isso é real"),
+        ev("post", "B", 2, "Fala sério que isso é real!"),
+      ];
+      const probe = runAllProbes({ events, agentIds: ["A", "B"], totalPulses: 2 })
+        .find(r => r.probe === "cross-agent-echo");
+      expect(probe).toBeDefined();
+      expect(probe!.measured).toBeGreaterThan(0);
+    });
+
     it("is wired into runAllProbes with a band", () => {
```
caiotheodoro, line 242: nit: two tests in this describe share the exact name "is wired into runAllProbes with a band" (this one and content-repetition's at line 196) — rename so failures are distinguishable.

caiotheodoro (commented):
One bug and one risk inline, both confirmed by running the probe against crafted transcripts; everything mechanical is green (clean merge onto post-#51 main, 855/855 tests, gate PASS, the PR-body bench row reproduces exactly). Both fixes are localized to the two new functions; after them this merges with no rebase needed.

matheusht:
@caiotheodoro friendly ping — band-wiring assertion added per review, all four tests green, audit PASS. Ready when you are.

Issue #33, opened 2026-08-20
stagnation_echo_chamber scenario scores anomalously low vs. the rest of the benchmark suite


caiotheodoro:
In the round-2 benchmark sweep, `stagnation_echo_chamber` scored the worst of any of the 16 scenarios: `in_character: 2`, `motive_authenticity: 1` — well below the ~3.5-4 mean across the rest of the suite. Every other stagnation-category scenario scored in the normal range, so this looks scenario-specific rather than a general pattern.

Needs a transcript read to find the actual cause — worth checking the scenario's seed events/agent configuration in `stagnation-attractors.ts` for something unusual (e.g. a seed state that puts an agent in a position the current prompt/persona setup handles poorly).

**Where to look**: `packages/shared/src/scenarios/stagnation-attractors.ts` (scenario definition), the saved transcript at `docs/eval/evidence/experiment-lowtemp-round2/scenarios/stagnation_echo_chamber.json` (local, not committed — regenerate via the benchmark script referenced in PR #22 if needed).

matheusht:
## Root cause (found + fixed)

Transcript read from an offline mock run shows the low scores were correct feedback about the **mock provider**, not the engine or scenario:

1. **React attractor**: the persona-aware mock's charged-react branch fires whenever affection/admiration ≥ 0.5 — which the echo-chamber seeds for every agent. Leo's emoji digest was keyed on agent only, so he posted identical 😂 reactions for 12+ consecutive pulses. A human doing that is out-of-character; the judge was right.
2. **Tone-mismatched replies**: Mariana's pack style examples skew conflict-flavored ("não me parece certo."), reading as non-sequiturs in an agreement-only room. That's persona-pack content → belongs to the style-examples refresh work, not this change.

## Fix shipped

React saturation cap (max two identical consecutive reacts per agent, non-react turns reset the streak) + pulse-salted emoji choice. Same transcript after: motive_authenticity 1→5, in_character 2→3 under the rule judge; expected signals still 100%. Regression test pins the cap; full analysis committed at docs/eval/evidence/echo-chamber-root-cause.md.

Closes #33 once shipped.

yoarajota:
[...]

PR #71, opened 2026-08-22
fix(eval): break the echo-chamber react attractor in the persona mock

matheusht:
## What

Fixes a mock-provider degeneracy that made one benchmark scenario read as wildly out-of-character:

- **Diagnosis** (offline transcript, committed as evidence): in rooms seeded with high affection/admiration, the persona-aware mock's charged-react branch fired every pulse for affected personas, and an agent-keyed-only emoji digest served the same emoji forever — one persona posted identical reactions for 12+ consecutive turns while another emitted conflict-flavored non-sequiturs. The scenario's worst-in-suite judged scores were *correct feedback about the mock*, not the engine.
- **Fix**: per-agent react saturation cap (max two identical consecutive emoji reacts; any non-react turn resets the streak) and pulse-salted emoji selection so choices vary within a scene.
- **Result**: on the same offline transcript, rule-judge `motive_authenticity` 1 → 5, `in_character` 2 → 3, expected signals still 100%.
- Regression test pins the two-identical-reacts cap across all four agents plus signal integrity; full root-cause writeup committed under `docs/eval/evidence/echo-chamber-root-cause.md`.
- Tone-mismatched reply pools are persona-pack content — deliberately left to the style-examples refresh work.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+2 regression tests)
- Before/after transcript metrics recorded in the committed evidence doc.


docs/eval/evidence/echo-chamber-root-cause.md
```diff
@@ -0,0 +1,53 @@
...
+On main, Leo — seeded with affection 0.9 / admiration 0.8 — commits
+`reaction_sent` on every eligible pulse and the emoji digest is keyed on
+agent id alone, so the sequence is one emoji repeated for the entire
+scene (20× the same reaction). Meanwhile Mariana's reply pool skews
+conflict-flavored ("não me parece certo.", "curioso. curioso mesmo."),
+reading as non-sequiturs in an agreement-only room.
+
+A persona who posts nothing but 😂 for twenty pulses is genuinely
+out-of-character: the low judged scores were correct feedback about the
+mock provider, not about the engine or the scenario definition.
```
caiotheodoro: risk: "the low judged scores were correct feedback about the mock provider" is mis-attributed — the round-2 transcript that produced #33's 2/1 scores was a qwen3:1.7b local-model run (mock provider not in the generation path) containing zero reaction events; it's dominated by schema fallbacks and repetition no-ops. The mock attractor is real and worth fixing, but it can't be what tanked those scores. Re-scope this paragraph and keep #33 open. (Also "meaningfulfulness" further down.)

packages/eval/src/test/echo-chamber-stability.test.ts
```diff
@@ -0,0 +1,43 @@
...
+ * emoji react loop (a dozen 😂 in a row), which read as wildly
+ * out-of-character and tanked judged authenticity.
+ */
+describe("persona-aware mock: echo-chamber stability", () => {
+  it("never emits more than two identical consecutive reactions per agent", async () => {
+    const scenario = getScenario("stagnation_echo_chamber");
+    if (!scenario) throw new Error("scenario missing from registry");
+    const artifact = await ScenarioRunner.run(scenario, { llmMode: "mock" });
+
+    const agents = ["caio", "leo", "mariana", "bruno"];
```
caiotheodoro: nit: the hardcoded roster `["caio","leo","mariana","bruno"]` silently thins coverage if the scenario cast changes — derive the agent list from `artifact.events` instead.

caiotheodoro (commented):
The fix is sound and I could reproduce everything: on main leo reacts 😂 twenty times straight, after the fix max streak is 1, the regression test fails on main and passes merged, and the full 123-scenario bench is bit-identical on signals/probes/judge means (calibration kappa 0.129 both sides — zero side effects). Two corrections needed before merge, one inline and one on the body: the "Result" bullet claims rule-judge motive_authenticity 1→5 and in_character 2→3, but measured axes are bit-identical before/after — your own evidence doc's later commits retract this; update the body to match. And don't frame this as resolving #33 (see inline). Merge is conflict-free onto current main.

matheusht:
@caiotheodoro ping — evidence doc re-scoped to exactly what reproduces offline (identical-emoji loop eliminated head-to-head; judged improvement deferred to live runs), streak-reset restored with accurate control-flow comment, unrelated doc edits removed from the branch. Three reviewer rounds, final APPROVE.

PR #72, opened 2026-08-22
feat(eval): named benchmark slices (--slice edges|golden|canary)

matheusht:
## What

Makes benchmark slices first-class so per-axis samples stay meaningful:

- New `bench-slices.ts` module with curated, validated id sets: `edges` (all four edge_chaos scenarios), `golden` (the calibration overlap), `canary` (the documented fast iteration loop).
- `bench` CLI gains `--slice <name>`, composable with existing `--scenarios/--category/--limit`; unknown names fail loudly with the available list.
- Registry-resolution + edge-coverage + defensive-copy tests; docs run commands updated.

## Why

An axis trend was recently being read off a single scenario because the slice was a hand-typed list that happened to include only one edge_chaos scene. Curated slices make the sample explicit and test-enforced.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+4)
- Live run: `--slice edges` → 12 variant runs across all four edge scenarios, 0 failed, signals 100%.


matheusht:
@caiotheodoro ping — golden slice is now DERIVED from GOLDEN_LABELS (cannot drift), --slice/--scenarios conflict throws loudly, unknown slice ids name themselves in the error, composition semantics documented.

PR #73, opened 2026-08-22
feat(agent): motive-anchor the repetition retry correction

matheusht:
## What

Hardens the repetition-guard retry correction against a subtle prompt failure: bare "say something genuinely different" can be read as license for authentic-sounding but unmotivated novelty — the exact mechanism suspected of nudging `motive_authenticity` downward.

- The correction note now anchors the alternative to the character's actual state: "…while staying true to what you actually want right now; do not invent novelty that your current motive and emotional state wouldn't justify." The no-op escape hatch is preserved verbatim.
- New contract test pins all three components of every retry correction — repeat warning with quoted offending text, motive anchor, escape hatch — so none can silently regress.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+1 contract test; existing repetition tests unchanged and green)
- Behavioral measurement (does judged motive authenticity move?) requires live-model runs via the sweep harness — documented as maintainer follow-up.


packages/server/src/agent/agent-runtime.ts
```diff
@@ -148,7 +148,7 @@ export class AgentRuntime {
       repetitionRetried = true;
       const retryPrompt = {
         ...prompt,
-        system: `${prompt.system}\n\nIMPORTANT: your last attempt this turn ("${intent.visibleContent}") was too close to something you already said. Say something genuinely different — a new angle, a reaction to someone else, a topic change — or choose "no_op" if you truly have nothing new to add.`,
+        system: `${prompt.system}\n\nIMPORTANT: your last attempt this turn ("${intent.visibleContent}") was too close to something you already said. Say something genuinely different — a new angle, a reaction to someone else, a topic change — while staying true to what you actually want right now; do not invent novelty that your current motive and emotional state wouldn't justify. Or choose "no_op" if you truly have nothing new to add.`,
```
caiotheodoro: bug: this line no longer exists on main — #51 moved the retry arm into `surface/action-intent-step.ts`, where the correction string appears twice (~141–148): once inside `promptVersionHash([...])` and once as `system:`. The rebase must re-apply the change to BOTH copies or the recorded promptVersion stops hashing the text actually sent. Extracting a single note-builder makes the desync impossible; the auto-merged contract test will catch a wrong resolution, which helps.

caiotheodoro: question: the title says motive-anchor, but nothing anchors on a concrete motive — `intent.privateMotiveSummary` is in scope right here and unused; "what you actually want right now" is a generic pointer the model must resolve back through the prompt sections. Given the failure population is small models that already ignore the no-repeat instruction, should the note quote the actual motive (accepting the risk that quoting the failed attempt's motive could re-entrench the loop), or should the PR retitle to what it is — a generic grounding clause?

caiotheodoro: risk: repetition often happens precisely because the motive is unchanged, so "do not invent novelty your current motive wouldn't justify" plausibly narrows the retry's escape space and pushes more second attempts back over the Jaccard threshold — trading repetition-spam for no_op silence (the no_op clause is now its own sentence right after a prohibition, which small models over-select). This lands unmeasured; #69's sweep (guard-block counts + cohesion judge) is exactly what would catch it.

packages/server/src/agent/__tests__/agent-runtime.test.ts
```diff
@@ -271,5 +271,29 @@ describe("AgentRuntime Orchestration", () => {
       expect(output.intent.privateMotiveSummary).toContain("even after a retry");
       expect(output.operatorEvents.some(e => e.type === "intent_blocked")).toBe(true);
     });
+
+    it("retry correction carries the repeat warning, a motive anchor, and the no-op escape hatch", async () => {
```
caiotheodoro, line 275: risk: this proves the words appear in the second request; the mock returns a canned different reply regardless, so nothing asserts any behavioral difference — the body's causal claim ("the exact mechanism suspected of nudging motive_authenticity downward") is untested by construction. Fine as a regression pin; frame the merge as a hypothesis-bearing tweak pending the sweep measurement.

packages/server/src/agent/__tests__/agent-runtime.test.ts
```diff
@@ -271,5 +271,29 @@ describe("AgentRuntime Orchestration", () => {
...
+      const retryBody = JSON.parse(fetchMock.mock.calls[1]![1]!.body as string);
+      const retrySystem: string = retryBody.messages[0]!.content;
+      // (a) repeat warning with the offending text quoted
+      expect(retrySystem).toContain("was too close to something you already said");
+      expect(retrySystem).toContain('"kkkkk não acredito nesse take"');
+      // (b) motive anchor — novelty must stay motivated by current state
+      expect(retrySystem).toContain("staying true to what you actually want right now");
+      expect(retrySystem).toContain("do not invent novelty");
+      // (c) escape hatch
+      expect(retrySystem).toContain('choose "no_op" if you truly have nothing new to add');
```
caiotheodoro: nit: the body says the no_op escape hatch is "preserved verbatim", but the clause changed from "— or choose" to a standalone "Or choose" sentence — and this assertion starts at exactly the substring that masks the rewording. Pin the real sentence or drop "verbatim" from the body.

caiotheodoro (commented):
One bug and two risks inline; the diff is clean and honestly scoped (no stray hunks). PR-wide: this shares its stale base and retry region with #69 — suggest #73 lands first (one-sentence port), #69 rebases over it, and both extract the correction-note into one shared builder so the hash/system duplication on main can't desync; #69's sweep harness is also the instrument this PR's body says it needs for behavioral evidence. The forced port is also the natural moment to promote the note to a fenced container per the merged #51 conventions (model output is currently interpolated unfenced into instruction text — pre-existing, not a regression here).

matheusht:
## Note on CI: GitHub-side runner stall

This branch's 7 check-suites have been stuck in `queued` with no app/runner attached (verified via API — `status: queued`, `conclusion: null`, no slug) across multiple pushes and a close/reopen cycle. Every other branch's `gate` workflow runs normally, so this looks like an Actions scheduling stall on this specific ref rather than a repository problem.

**Local verification of the same merge ref** (full gates):

- `pnpm -r build` ✅
- `pnpm lint` ✅ (exit 0)
- `pnpm test:all` ✅ — all suites pass, including the new retry-correction contract test
- `scripts/audit-tests.mjs --check` ✅ PASS, zero violations
- Independent reviewer verdict: **APPROVE**, zero findings

If a maintainer re-runs the workflow or pushes any commit to this branch, the queued suites should either execute or clear. The change itself is a two-hunk prompt-string hardening plus one contract test — low risk.

matheusht:
@caiotheodoro ping — heads-up: this branch's check-suites are stuck queued GitHub-side (no runner attached across pushes + close/reopen; other PRs run fine). Local gates are fully green and the reviewer verdict was APPROVE with zero findings — details in the earlier comment. A manual workflow re-run or any push should clear it.

Issue #36, opened 2026-08-20
Re-evaluate qwen3:8b now that channelType/repetition fixes exist


caiotheodoro:
This session ran a qwen3:8b-vs-1.7b comparison and found 8b 5-6x slower with no clear quality win (one scenario scored *worse* on 8b, uniform 3s vs 1.7b's mostly-4s; another was a near-tie). But that comparison has confounds worth re-examining:

- It happened *before* the repetition-retry logic (PR #22) existed, so 8b's extra token cost from retries wasn't isolated from its baseline generation cost.
- Per-scenario timing escalated within the same run (306s → 567s → 672s for consecutive scenarios), suggesting something compounding across the run (context length growth? thermal throttling on the test machine?) rather than a clean per-scenario cost comparison.

Worth a cleaner re-run: fixed context length, same machine at a consistent thermal state, with the current (post-PR #22) prompt/validation logic, comparing 1.7b vs 8b on identical scenarios with identical seeds.

**Where to look**: the benchmark script pattern from PR #22 (see PR description for the `run-experiment.mjs`-equivalent approach), `examples/simulations/qwen3-8b.example.json`.

yoarajota:
Addressed by PR #74: the confound-controlled re-run protocol is at docs/eval/qwen3-comparison-protocol.md (interleaved counterbalanced arms, per-scenario restarts, seed + multi-seed sweeps with Ollama version/digest fingerprint, cross-family judge instrument, full-point/multi-axes decision rule). The live re-run itself is a maintainer task per that doc.

Issue #28, opened 2026-08-20
LLM judge is the same model family as generation (self-preference bias risk)


caiotheodoro:
In every local benchmark run this session, the judge and the generating model were the same model (`qwen3:1.7b` judging `qwen3:1.7b`-generated transcripts). Standard LLM-eval practice flags same-family judge/generator pairing as a self-preference bias risk — the judge may score the generator's characteristic output patterns more favorably than an independent judge would, inflating or skewing axis scores in ways that are hard to detect from the numbers alone.

Worth an experiment: run the same transcripts through a differently-sourced judge (e.g. a larger model, or a different model family entirely) and compare axis scores. If they diverge meaningfully, that's evidence the current judge setup needs a fix (different default judge model, or a "jury of judges" approach — majority verdict across 3+ independently-sourced judges, which the LLM-eval literature shows measurably reduces this bias).

**Where to look**: `packages/eval/src/judge/judge.ts`'s `llmJudge`/`LlmJudgeConfig`, `docs/eval/README.md`.

matheusht:
## Plan

1. New `juryJudge(scenario, events, configs[])` in the judge module: runs the existing `llmJudge` once per independently-sourced judge config (different model/family/endpoint), then aggregates per-axis by MEDIAN — the ordinal-appropriate majority verdict that measurably reduces self-preference bias. A judge that errors doesn't poison the verdict (dropped; needs ≥1 survivor). Also returns per-judge scores for divergence inspection: meaningful spread between judges IS the bias evidence this issue asks to detect.
2. Unit-tested offline with fetch mocks: three agreeing judges → identical medians; one biased outlier → median resists; two of three failing → still verdicts; all failing → throws honestly.
3. Live comparison protocol documented in docs/eval/README.md: how to point 2+ differently-sourced endpoints at the same transcripts (`PERFECTMAN_JUDGE_*` envs already parameterize model/baseUrl) and read divergence from the returned per-judge scores. Actual cross-family runs are maintainer-executed.

Closes #28 once shipped.

yoarajota:
[...]

PR #74, opened 2026-08-22
docs(eval): controlled qwen3 1.7b-vs-8b comparison protocol

matheusht:
## What

Adds a written re-run recipe for the qwen3 1.7b-vs-8b question that removes the three confounds from the first comparison:

- **Retry cost isolated**: both arms run on current main and record wire-call counts alongside turn counts, so retry pressure is visible separately from baseline generation cost.
- **Thermal/KV compounding controlled**: models interleaved per scenario instead of back-to-back arms, first scenario discarded as warm-up, server restarted between scenarios.
- **Sampling pinned**: identical seed across arms by default.

Includes an explicit pre-registered decision rule (8b must win ≥ two quality axes by a full point at ≤2× latency with comparable retry pressure — anything narrower is noise at n=4) so the read-out can't be post-hoc rationalized. All model-touching steps are marked maintainer-run; slice selection and dry-runs work fully offline via the named-slices CLI.

## Validation

Docs-only. `pnpm lint` PASS - `pnpm test:all` PASS (850).


packages/eval/src/bench-slices.ts
```diff
@@ -0,0 +1,29 @@
...
+  ],
+  /** The golden-labeled set — derived so it cannot drift from calibration. */
+  get golden(): readonly string[] {
+    return GOLDEN_LABELS.map(g => g.scenarioId);
+  },
+  /** The documented iteration canary — fast before/after loop. */
+  canary: ["motive_gossip", "v1_exclusion_inferred", "motive_conflict", "stagnation_resentment_loop"],
+};
+
+export function resolveBenchSlice(name: string): string[] | undefined {
```
caiotheodoro, line 31: bug: this file (plus the --slice hunks in cli/bench.ts and bench-slices.test.ts) duplicates already-merged #72 at the pre-fix revision — missing main's 2ab503f: the `Object.hasOwn` guard, the --slice/--category mutual exclusion, the empty-slice guard, and the bare-`--slice` fallthrough fix. Rebase and drop all three code files (and both test-inventory files) — main already has newer versions.

docs/README.md
```diff
@@ -227,17 +227,15 @@ pnpm freellm:dev
 
 The script builds the FreeLLMAPI container from `https://github.com/tashfeenahmed/freellmapi`, reads this repo's `.env`, exposes the API on `http://localhost:3001/v1`, and exposes the dashboard on `http://localhost:5173`. Set `FREELLMAPI_ENCRYPTION_KEY` in `.env` before first run so FreeLLMAPI can store provider keys. Use the dashboard to add provider keys and create the unified key, then store that value in `.env` as `FREELLMAPI_KEY`.
 
-For the local Qwen3 config, run Ollama first. On macOS, prefer native
-Ollama (`brew install ollama && ollama pull qwen3:1.7b && ollama serve`) —
-Docker Desktop has no GPU passthrough there and the CPU-only container is
-dramatically slower. On Linux machines with an NVIDIA GPU, use the Docker
-path:
+For the DeepSeek config, copy `examples/simulations/deepseek.example.json` to `config/index.json` and set `DEEPSEEK_API_KEY` in your environment. DeepSeek is routed through the OpenAI-compatible provider using `https://api.deepseek.com/v1`.
```
caiotheodoro: bug: `examples/simulations/deepseek.example.json` doesn't exist on main and this PR doesn't add it — broken instruction.

docs/README.md
```diff
@@ -227,17 +227,15 @@ pnpm freellm:dev
...
+
+For the local Qwen3 config, run the Ollama container first:
```
caiotheodoro: bug: this reverts main's macOS-native-Ollama paragraph back to `pnpm qwen:dev` — a stale-base rewrite that also contradicts this PR's own protocol: the restart step (`pkill -f "ollama serve"` + wait for port 11434) only works against native Ollama; against the Docker container this points to, pkill matches nothing and the lsof loop spins forever.

docs/eval/qwen3-comparison-protocol.md
```diff
@@ -0,0 +1,66 @@
...
+
+A re-run recipe for the `qwen3:8b` vs `qwen3:1.7b` question. The first
+comparison (8b 5–6× slower, no clear quality win) carried three confounds
+this protocol removes.
+
+## Confounds being controlled
+
+| Confound | Control |
+| --- | --- |
+| Retry logic landed after the comparison; 8b's retry-token cost was folded into its baseline | Both arms run on current main (repetition guard included); per-scenario `fallbackCount` from the bench report is the retry-pressure proxy (wire-call-level accounting lands with the repetition-sweep work) |
```
caiotheodoro: risk: confound 1 isn't made visible, only equalized — `fallbackCount` counts terminal llm_failure events (scenario-runner.ts:173), so a call that succeeds on retry 2 emits nothing and retry cost stays folded into latencyMs, which is the confound #36 raised. The body's "record wire-call counts alongside turn counts" describes machinery that doesn't exist on main. Say equalized, not visible, until the wire-call accounting lands.

docs/eval/qwen3-comparison-protocol.md
```diff
@@ -0,0 +1,66 @@
...
+- judge axis means vs targets (quality),
+- `fallbackCount` per arm (retry pressure proxy),
+- wall-clock `latencyMs` compared within the pair, never across scenarios,
+- signals: track pass rate directionally (live-model runs legitimately
+  vary — the offline 100% bar is not the live bar), but any outright
+  scenario failure invalidates that pair.
+
+## Decision rule
+
+Prefer 1.7b unless 8b wins ≥ two quality axes by a full point across the
```
caiotheodoro, line 69: risk: the decision rule (8b wins ≥2 quality axes by a full point) is measured with `--judge rule`, whose axes main's own docs say not to trust — they're structural heuristics largely insensitive to prose quality, so a genuine 8b quality win may be unmeasurable by the chosen instrument and the default ("prefer 1.7b") wins by construction. The repo has the right tool: the DeepSeek judge at temperature 0 — also cross-family, sidestepping the #28 bias that local-qwen-as-judge would reintroduce (it would judge its own arms). Use it or justify not.

docs/eval/qwen3-comparison-protocol.md
```diff
@@ -0,0 +1,66 @@
...
+comparison (8b 5–6× slower, no clear quality win) carried three confounds
+this protocol removes.
+
+## Confounds being controlled
+
+| Confound | Control |
+| --- | --- |
+| Retry logic landed after the comparison; 8b's retry-token cost was folded into its baseline | Both arms run on current main (repetition guard included); per-scenario `fallbackCount` from the bench report is the retry-pressure proxy (wire-call-level accounting lands with the repetition-sweep work) |
+| Per-scenario timing escalated within a single run (306s → 567s → 672s) | Arms are interleaved **per scenario** (1.7b then 8b on the same scenario before moving on), the first pair is discarded as warm-up, and the Ollama server is restarted between scenarios to reset KV cache and thermal state |
+| Unpinned sampling made quality deltas unattributable | Local-mode benchmarks pin sampling by default (seed 42; see docs/eval README). Keep `PERFECTMAN_LLM_SEED` identical across arms |
```
caiotheodoro: risk: pinned by a single seed but no environment fingerprint — Ollama seed determinism is best-effort, format+seed is reported nondeterministic (ollama#12559), and the think:false silent-drop is version-dependent, so two Ollama versions can fail differentially by model. Require `ollama --version` + model digests in each report, and with n=3 pairs after the warm-up discard, sweep 2–3 seeds per arm so the full-point threshold means something.

docs/eval/qwen3-comparison-protocol.md
```diff
@@ -0,0 +1,66 @@
...
+  --scenarios "$SCENARIO" --out "out/qwen-8b-$SCENARIO.json"
+```
+
+Slice identity: these are exactly the `canary` slice's four scenarios
+(see `packages/eval/src/bench-slices.ts`) — the quality-focused
+iteration set both arms must share.
+
+## Read-out
+
+Per scenario pair (discard the first pair as warm-up):
```
caiotheodoro: question: the warm-up discard and the restart policy fight each other — the server restarts before every scenario, so each pair is equally cold; what residual effect is discarding the first pair (always motive_gossip) controlling for? Also arm order is never counterbalanced: 1.7b always runs first on a fresh server, 8b second with 1.7b still resident. Alternating per scenario costs nothing.

docs/eval/qwen3-comparison-protocol.md
```diff
@@ -0,0 +1,66 @@
...
+| Unpinned sampling made quality deltas unattributable | Local-mode benchmarks pin sampling by default (seed 42; see docs/eval README). Keep `PERFECTMAN_LLM_SEED` identical across arms |
+
+## Recipe
+
+One scenario per invocation — that is what makes the interleave and the
+restarts possible:
+
+```sh
+# repeat for SCENARIO in motive_gossip v1_exclusion_inferred motive_conflict stagnation_resentment_loop
+SCENARIO=motive_gossip
```
caiotheodoro, line 27: nit: the recipe never pulls the models — a fresh `ollama serve` fails the first call for whichever tag is missing. Add `ollama pull qwen3:1.7b qwen3:8b` as a prereq, and note `--out` lands under packages/eval/ because `pnpm --filter` sets the cwd.

docs/eval/qwen3-comparison-protocol.md
```diff
@@ -0,0 +1,66 @@
...
+
+Slice identity: these are exactly the `canary` slice's four scenarios
+(see `packages/eval/src/bench-slices.ts`) — the quality-focused
+iteration set both arms must share.
+
+## Read-out
+
+Per scenario pair (discard the first pair as warm-up):
+
+- judge axis means vs targets (quality),
```
caiotheodoro: nit: the report already carries promptVersions/templateVersions from #51 — require they be identical across both arms of every pair as a free sanity check that the arms ran the same prompt stack.

caiotheodoro (commented):
Three bugs and three risks inline. The protocol doc itself is well-grounded — every flag, env var, and report field it cites exists on main, the canary ids match exactly, and #36's numbers are quoted faithfully. But the PR isn't docs-only: it ships stale duplicates of merged #72 at the pre-2ab503f revision, and that's the conflict — a head-favoring resolution reverts four --slice hardening fixes. Correct resolution: take main for everything except qwen3-comparison-protocol.md and the two README link additions, which is all the genuinely new content.

matheusht:
@caiotheodoro ping — recipe now implements its own controls (per-scenario interleave with pkill/relaunch restarts), fallbackCount-only retry proxy (turn-count claim dropped), canary identity fixed, port-free readiness poll added.

Issue #44, opened 2026-08-20
No temperature-sensitivity study — all narrative-quality benchmarking used a single fixed value (0.25)


caiotheodoro:
Every benchmark run this session (PRs #19-#22, the before/after narrative-quality numbers) used a single fixed generation temperature of 0.25 ("low temperature" per the original request). That value was never compared against alternatives — it's unknown whether 0.25 is actually the right setting, or whether it's part of *why* `creativity_unhinged` (#29) and `unpredictability` (#32) are the weakest/flattest axes.

Worth a dedicated sweep: run the same canary scenario set (see PR #22's 4-scenario canary pattern) at several temperatures (e.g. 0.25 / 0.5 / 0.7 / 1.0) and compare axis scores *together with* the repetition-guard's trigger rate (`repetition-guard.ts`, #35) — the expected tradeoff is higher temperature should raise creativity/unpredictability but also raise raw repetition-guard trigger rate before any retry recovers it, and coherence axes may drop. Right now that tradeoff curve is completely unmeasured; the current 0.25 default has no evidence behind it beyond "low temperature was the original ask."

Blocked/complicated by unseeded generation (see the reproducibility issue also opened this batch) — without a fixed seed, temperature comparisons are confounded by ordinary sampling variance, not just the temperature change itself.

**Where to look**: `packages/server/src/agent/persona-loader.ts` (`sampling.temperature` per persona pack), the benchmark script pattern from PR #22.

matheusht:
## Plan

Same harness pattern as the repetition sweep, one axis over:

1. New offline CLI `sweep:temperature`: grid [0.25, 0.5, 0.7, 1.0] × the canary scenario set, seed pinned (42) per the reproducibility fix — each cell clones the persona packs with `sampling.temperature` overridden and runs through a custom provider factory.
2. Per-cell metrics from real artifacts: rule-judge axis means (creativity_unhinged, narrative_cohesion, unpredictability where rubric provides), repetition-guard block count (the expected tradeoff curve), content-repetition probe mean.
3. Committed mock-mode evidence with honest scoping: the deterministic mock's only temperature sensitivity is its charged-react gate at ≥ 0.9, so sub-0.9 cells are expected identical — the harness exists so live-model sweeps drop straight in; that protocol is the maintainer follow-up.

Closes #44 once shipped.

yoarajota:
[...]

PR #75, opened 2026-08-22
feat(eval): generation-temperature sweep harness with pinned seed

matheusht:
## What

Measures the generation-temperature tradeoff that the 0.25 default never had:

- New offline CLI `pnpm --filter @perfectman/eval sweep:temperature`: grid [0.25, 0.5, 0.7, 1.0] × the canary scenario set, seed pinned (42) so cells differ only by temperature — the confound the reproducibility work removed.
- Each cell clones the persona packs with `sampling.temperature` overridden and runs through a custom provider factory; per-cell metrics from real artifacts: repetition-guard block count (the expected cost curve), content-repetition probe mean, and rule-judge axis means.
- Committed mock evidence already shows the expected shape: 0.25–0.7 identical (the deterministic mock's only sensitivity is its charged-react gate at ≥ 0.9), while 1.0 diverges — fewer guard blocks, lower content repetition, higher creativity. Honest scoping included; live-model sweeps drop straight into this harness.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS
- Sweep executed end-to-end offline; 4-cell matrix committed under docs/eval/evidence/.

### DIFF
--- a/docs/eval/evidence/temperature-sweep-mock.json
+++ b/docs/eval/evidence/temperature-sweep-mock.json
@@ -1,0 +1,77 @@
+{
+  "version": "temperature-sweep-v1",
+  "mode": "mock",
+  "generatedAt": "2026-08-22T12:14:53.583Z",
+  "scenarios": [
+    "motive_gossip",
+    "v1_exclusion_inferred",
+    "motive_conflict",
+    "stagnation_resentment_loop"
+  ],
+  "grid": [
+    {
+      "temperature": 0.25,
+      "runs": 4,
+      "guardBlocks": 92,
+      "contentRepetitionMean": 0.302,
+      "axisMeans": {
+        "in_character": 3.25,
+        "voice_match": 2.25,
+        "motive_authenticity": 5,
+        "interpretation": 5,
+        "creativity_unhinged": 4,
+        "memory_continuity": 4,
+        "no_ai_leak": 5,
+        "narrative_cohesion": 5
+      }
+    },
+    {
+      "temperature": 0.5,
+      "runs": 4,
+      "guardBlocks": 92,
+      "contentRepetitionMean": 0.302,
+      "axisMeans": {
+        "in_character": 3.25,
+        "voice_match": 2.25,
+        "motive_authenticity": 5,
+        "interpretation": 5,
+        "creativity_unhinged": 4,
+        "memory_continuity": 4,
+        "no_ai_leak": 5,
+        "narrative_cohesion": 5
+      }
+    },
+    {
+      "temperature": 0.7,
+      "runs": 4,
+      "guardBlocks": 92,
+      "contentRepetitionMean": 0.302,
+      "axisMeans": {
+        "in_character": 3.25,
+        "voice_match": 2.25,
+        "motive_authenticity": 5,
+        "interpretation": 5,
+        "creativity_unhinged": 4,
+        "memory_continuity": 4,
+        "no_ai_leak": 5,
+        "narrative_cohesion": 5
+      }
+    },
+    {
+      "temperature": 1,
+      "runs": 4,
+      "guardBlocks": 44,
+      "contentRepetitionMean": 0.078,
+      "axisMeans": {
+        "in_character": 3.75,
+        "voice_match": 2.75,
+        "motive_authenticity": 5,
+        "interpretation": 5,
+        "creativity_unhinged": 4.75,
+        "memory_continuity": 4,
+        "no_ai_leak": 5,
+        "narrative_cohesion": 5
+      }
+    }
+  ]
+}
--- a/docs/eval/evidence/temperature-sweep-mock.json
+++ b/docs/eval/evidence/temperature-sweep-mock.json
@@ -1,7 +1,7 @@
 {
   "version": "temperature-sweep-v1",
   "mode": "mock",
-  "generatedAt": "2026-08-22T12:14:53.583Z",
+  "slice": "canary",
   "scenarios": [
     "motive_gossip",
     "v1_exclusion_inferred",
--- a/packages/eval/package.json
+++ b/packages/eval/package.json
@@ -19,7 +19,8 @@
     "calibrate": "node dist/cli/calibrate.js",
     "narrate": "node dist/cli/narrate.js",
     "test": "vitest run",
-    "typecheck": "tsc -p tsconfig.json --noEmit"
+    "typecheck": "tsc -p tsconfig.json --noEmit",
+    "sweep:temperature": "node dist/cli/sweep-temperature.js"
   },
   "dependencies": {
     "@perfectman/engine": "workspace:*",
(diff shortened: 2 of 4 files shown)


packages/eval/src/cli/sweep-temperature.ts
```diff
@@ -0,0 +1,184 @@
...
+): RoleplayScenario[] {
+  if (!ids || ids.length === 0) {
+    throw new Error(`Slice "${SLICE}" resolved to no scenarios — nothing to sweep`);
+  }
+  return ids.map(id => {
+    const scenario = getScenario(id);
+    if (!scenario) throw new Error(`Slice "${SLICE}" references unknown scenario "${id}"`);
+    return scenario;
+  });
+}
```
caiotheodoro, line 78: bug: this harness structurally can't run the study #44 asks for. `llmMode: "mock"` is hardcoded (no `--mode local`, unlike the sibling evidence CLI), and deeper: the temperature override works by cloning packs into the providerFactory — a path that only exists for mocks. In local mode `scenarioToConfig` calls `getPersonaPackById` itself and builds `localLLMConfig` from the canonical pack's temperature; the cloned packs never reach it. A live sweep needs a `samplingOverride` in RunnerOpts threaded into `localLLMConfig` first. Either add that (plus the expected live wall-clock cost) or rescope the PR body to "sweep plumbing, mock-only; live path TODO on #44".

packages/eval/src/cli/sweep-temperature.ts
```diff
@@ -0,0 +1,184 @@
...
+
+function packsAtTemperature(temperature: number): Map<string, PersonaPack> {
+  const packs = new Map<string, PersonaPack>();
+  for (const persona of ALL_PERSONAS) {
+    const pack = getPersonaPackById(persona.id);
+    if (pack) {
```
caiotheodoro, line 84: bug: the generic-pack fallback is dead code with a false comment — `getPersonaPackById("generic")` always returns undefined (no pack has that id), so the real fallback is Goulart's pack, and the default factory it claims to match actually falls back to an inline synthetic pack. Unreachable today only because all four canary scenarios have compiled packs. Export the factory's generic pack and reuse it with temperature overridden, or delete the branch and fail loudly.

docs/eval/evidence/temperature-sweep-mock.json
```diff
@@ -0,0 +1,77 @@
+{
+  "version": "temperature-sweep-v1",
+  "mode": "mock",
```
caiotheodoro, line 3: risk: three of four cells are identical by construction (the mock's only temperature read is the ≥0.9 charged-react gate), so the 1.0-row divergence is a hand-authored branch, not signal — and its guardBlocks drop (92→44) points opposite to #44's hypothesis, so a skimmer takes away an inverted conclusion. Embed a note in the JSON saying it's a plumbing fixture and the 1.0 divergence is a mock artifact, and fix the PR body's "expected cost curve" framing.

packages/eval/src/cli/sweep-temperature.ts
```diff
@@ -0,0 +1,184 @@
...
+      guardBlocks: 0,
+      contentRepetitionMean: 0,
+      axisMeans: {},
+    };
+    const axisSums: Record<string, { sum: number; count: number }> = {};
+    let repetitionSum = 0;
+    let repetitionRuns = 0;
+
+    for (const scenario of scenarios) {
+      const artifact = await ScenarioRunner.run(scenario, {
```
caiotheodoro, line 122: risk: the report omits every attribution field the repo's reproducibility work established — no seeds, no model, no promptVersions/templateVersions (computed per-run on the artifact and discarded at L90-105), no provider fingerprint. Also the body's "seed pinned (42)" is wrong: benchSeed only feeds local mode, which this never uses; the mock is seeded from `scenario.seed`. Record the versions/model per cell, say what actually pins the seeds, and reserve fingerprint fields for local mode — this file is the template live sweeps inherit.

packages/eval/src/cli/sweep-temperature.ts
```diff
@@ -0,0 +1,184 @@
...
+ * The prefix `ActionIntentStep` stamps on a repetition-guard fallback's
+ * `privateMotiveSummary`. `guardBlocks` is the sweep's headline metric and
+ * this prose is its only signal, so rewording that sentence upstream would
+ * zero the metric with no other symptom — `sweep-temperature.test.ts` pins
+ * the marker against a real run so the reword fails loudly instead.
+ */
+export const REPETITION_GUARD_MOTIVE_MARKER = "Repetition guard";
+
+type Cell = {
+  temperature: number;
```
caiotheodoro, line 56: nit: `countGuardBlocks` hardcodes the "Repetition guard" substring; #69 exports REPETITION_GUARD_MARKER and ships a near-identical helper. Whichever lands second should reuse it, or a guard-message rewording silently zeroes this metric.

packages/eval/src/cli/sweep-temperature.ts
```diff
@@ -0,0 +1,184 @@
...
+import { writeFileSync, mkdirSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import type { PersonaPack, RoleplayScenario } from "@perfectman/shared";
+import { getScenario, getPersonaPackById, ALL_PERSONAS } from "@perfectman/shared";
+import { ScenarioRunner } from "../run/scenario-runner.js";
+import { ruleJudge } from "../judge/judge.js";
+import { personaAwareProviderFactory } from "../bench/persona-aware-mock.js";
+import { resolveBenchSlice } from "../bench-slices.js";
+
+const args = process.argv.slice(2);
```
caiotheodoro, line 30: nit: the canary ids are retyped — `resolveBenchSlice("canary")` from merged #72 exists so no axis hangs on a hand-typed list again. Also report the post-filter ids, not the raw SCENARIOS (L61 silently drops unresolvable ones).

packages/eval/package.json
```diff
@@ -19,7 +19,8 @@
     "calibrate": "node dist/cli/calibrate.js",
     "narrate": "node dist/cli/narrate.js",
     "test": "vitest run",
-    "typecheck": "tsc -p tsconfig.json --noEmit"
+    "typecheck": "tsc -p tsconfig.json --noEmit",
+    "sweep:temperature": "node dist/cli/sweep-temperature.js"
```
caiotheodoro, line 23: nit: nothing points at the new script — one line each in docs/eval/README.md and the evidence README's artifact table.

caiotheodoro (commented):
Two bugs and two risks inline. The skeleton is sound and the committed mock matrix reproduces byte-for-byte (ran it: 0.7s, identical modulo timestamp). But the central promise — "live-model sweeps drop straight into this harness" — is false twice over, and the mock evidence narrates an if-branch as a cost curve, in the opposite direction from #44's hypothesis. Coordination note: #69 ships a near-identical countGuardBlocks and REPETITION_GUARD_MARKER; whichever lands second should dedup.

Issue #29, opened 2026-08-20
creativity_unhinged is the weakest axis and barely moved (2.40 → 2.73)


caiotheodoro:
Across a 16-scenario benchmark sweep, `creativity_unhinged` went from the weakest baseline axis (2.40/5) to still the weakest after a round of fixes (2.73/5, +0.33) — every other targeted axis moved more (`narrative_cohesion` +0.80, `memory_continuity` +0.74, `no_ai_leak` +0.68). None of this session's fixes directly targeted creativity/unhinged-ness, so this isn't surprising, but it's now the clearest remaining gap.

Worth investigating whether persona `styleExamples`/`voiceGuidelines` in the persona packs are structurally capping creative range (e.g. too safe/generic example phrasing), independent of the schema/repetition bugs already fixed in PRs #19-#22.

**Where to look**: `packages/shared/src/persona/` (persona pack definitions), `packages/server/src/agent/action-intent-prompt-builder.ts`'s `renderStyleExamples` (note: broader prompt-builder restructuring is already claimed by another contributor — see repo discussion — scope this to persona *content*, not prompt *structure*), `packages/eval/src/judge/rubrics.ts` for the axis's scoring anchors.

yoarajota:
Resolved by post-issue work on main: `creativity_unhinged` now clears its 3.5 target — 4.081 in the pinned mock baseline (`docs/eval/bench-mock-baseline.json`) and 4.086 in `docs/eval/evidence/evidence-report.md`. `docs/eval/lora-feasibility-spike.md` (2026-08-22) explicitly classifies it as no longer a weak axis: 'currently `creativity_unhinged` already clears its target there, so gains there are a bonus, not the motivation.' Contributing merges: intent-entropy/unpredictability (#79), persona-style refresh (#78). Closing.

PR #76, opened 2026-08-22
docs: LoRA per-persona fine-tuning feasibility spike

matheusht:
## What

Adds a feasibility spike document for per-persona LoRA fine-tuning as an alternative to prompt-stuffing:

- **Inventory of what already exists**: compiled persona packs carry style examples, voice patterns, motive lexicons, and memory seeds — already in the 10–20-curated-examples range research suggests is sufficient for fixed-persona SLM dialogue; the local Ollama infra supports `ADAPTER` layers.
- **Experiment shape**: JSONL export from past transcripts → one adapter per persona on qwen3:1.7b → Modelfile registration → A/B benchmark (adapter + trimmed prompt vs base + full prompt) over the named slices with pinned sampling. Maintainer-run by design (live models); everything up to the invocations is offline-reproducible.
- **Pre-registered decision rule**: proceed only if `voice_match` gains ≥ 0.5 on average, `interpretation` does not regress (the documented cost of role-specific fine-tuning is weakened general reasoning), and prompt-token savings ≥ 30% of the persona section.
- **Known risks** recorded: commonsense forgetting, adapter/version drift, serving-matrix growth at five personas, unverified annotation-free alignment claims.

No runtime code touched.

## Validation

Docs-only. `pnpm lint` PASS. All referenced paths verified to exist.


docs/lora-feasibility-spike.md
```diff
@@ -0,0 +1,61 @@
...
+the local Qwen3 models could move that identity into the weights: fewer
+prompt tokens per call, potentially stronger `voice_match` / `in_character`
+/ `creativity_unhinged` (the persistently weakest axes).
+
+## What already exists (no new collection needed)
+
+- **Seed data**: compiled persona packs carry style examples, voice
+  patterns, private-motive lexicons, and memory seeds per persona
+  (`packages/shared/src/persona-packs/`). Research suggests 10–20 curated
+  examples per persona can be enough for fixed-persona SLM dialogue — the
```
caiotheodoro: bug: "(no new collection needed)" fails twice. The source this traces to (arXiv 2511.10277, Fixed-Persona SLMs) uses 10–20 handcrafted prompt–response pairs only as SEEDS for a synthetic-expansion pipeline that trains on ~115–564 reviewed pairs — so a data-prep step IS new collection (https://arxiv.org/html/2511.10277). And the packs aren't in range anyway: styleExamples counts on main are bruno 9, caio 9, goulart 9, leo 10, mariana 9 — single one-line utterances, not prompt–response pairs. Retitle the section and add a data-preparation step to the experiment shape.

docs/lora-feasibility-spike.md
```diff
@@ -0,0 +1,61 @@
...
+  packs are in that range today.
+- **Serving path**: Ollama supports Modelfiles with `ADAPTER` LoRA layers;
```
caiotheodoro: risk: "Ollama supports Modelfiles with ADAPTER LoRA layers" hides the constraint that matters for this stack — safetensors adapter import supports Llama/Mistral/Gemma only, not Qwen (https://docs.ollama.com/modelfile; ollama's convert/ has no qwen adapter converter). A PEFT LoRA for qwen3:1.7b needs a llama.cpp convert_lora_to_gguf.py detour, docs warn "erratic" behavior unless trained on the exact base, and the single-adapter/no-hot-swap limit is still open (ollama#14032, #9548) — five personas means five separately created and loaded models. Add a step 0: smoke-test a trivial GGUF adapter on qwen3:1.7b under the pinned Ollama version before spending anything on training.

docs/lora-feasibility-spike.md
```diff
@@ -0,0 +1,61 @@
...
+  stack the benchmarks already run on.
+- **Measurement**: the bench harness + named slices give before/after axis
+  scores with pinned sampling; the repetition guard and echo probes measure
+  whether fine-tuned personas repeat or converge differently.
+
+## Experiment shape (maintainer-run; needs live models)
+
+1. Export each persona pack to a JSONL chat dataset: system = compiled
+   identity frame; user turns from past scenario transcripts where that
+   persona acted; assistant turns = the persona's actual committed content.
```
caiotheodoro: risk: training-data circularity — the only persona-committed transcripts on main are outputs of exactly the pipeline whose weak voice motivated this spike (rule-judge voice_match mean 2.686; #29 still weakest axis). Fine-tuning on those turns is self-distillation of the current voice and can't be expected to lift voice_match above its source. Require filtering to high-judge-scored turns or curated/synthetic data — which again contradicts the no-new-collection heading.

docs/lora-feasibility-spike.md
```diff
@@ -0,0 +1,61 @@
...
+   Start with the two strongest-voice personas only.
+2. Train one LoRA adapter per persona on qwen3:1.7b (small model first —
+   cheapest signal). Keep rank ≤ 16, one epoch, then stop.
+3. Register a `Modelfile.perfectman-<persona>` with `ADAPTER`.
+4. Benchmark A/B: base model + full prompt vs adapter + trimmed prompt
+   (identity section removed), same slice (`--slice canary`), same seed.
```
caiotheodoro: risk: the A/B changes two variables at once (adapter added AND identity section removed), so a voice_match delta is unattributable. The base-model + trimmed-prompt arm alone costs zero training and establishes today how much the persona section buys on the weak axes — if removing it barely moves voice_match, the whole premise changes. Run it as a pre-experiment.

docs/lora-feasibility-spike.md
```diff
@@ -0,0 +1,61 @@
...
+5. Measure: the three target axes AND `interpretation` — the known cost of
+   role-specific fine-tuning is weakened general reasoning, so a rise in
+   voice scores that coincides with an `interpretation` drop is a net loss.
+
+## Decision rule
+
+Proceed further only if: `voice_match` improves by ≥ 0.5 on average across
+trained personas, `interpretation` does not fall below its current mean,
+and prompt-token savings ≥ 30% of the persona section. Otherwise stay
```
caiotheodoro: nit: "prompt-token savings ≥ 30% of the persona section" isn't an experimental outcome — step 4 fixes the trim by construction, so the number is computable today from renderPersona. Compute it now, or restate as something falsifiable (tokens removed at non-regressing quality).

docs/lora-feasibility-spike.md
```diff
@@ -0,0 +1,61 @@
...
+- **Seed data**: compiled persona packs carry style examples, voice
+  patterns, private-motive lexicons, and memory seeds per persona
+  (`packages/shared/src/persona-packs/`). Research suggests 10–20 curated
+  examples per persona can be enough for fixed-persona SLM dialogue — the
+  packs are in that range today.
+- **Serving path**: Ollama supports Modelfiles with `ADAPTER` LoRA layers;
+  the docker/native infra under `docker/qwen3/` is the same free/local
+  stack the benchmarks already run on.
+- **Measurement**: the bench harness + named slices give before/after axis
+  scores with pinned sampling; the repetition guard and echo probes measure
```
caiotheodoro: nit: there's no probe named "echo" — the bench probe is `content-repetition` (contentRepetitionRate); "echo" names a scenario and a mock-stability test. Name the real probe so a future reader can find it.

caiotheodoro (commented):
One bug and three risks inline. Credit where due: the pre-registered decision rule, forgetting guard, and honest unknowns section are the right skeleton, and every repo path the doc cites checks out. But both load-bearing feasibility claims are wrong as written, with sources inline. One PR-wide gap: the spike argues for training infrastructure without weighing the cheaper levers already in-repo for the same axes — #29's styleExamples-curation hypothesis, the packs' per-persona sampling knobs, and just-merged #68's salience retrieval — a spike proposing training should say why those don't cover the gap first.

docs/lora-feasibility-spike.md
caiotheodoro: The 17:36 commit softened the body but the heading still says "(no new collection needed)" while step 1 now describes building a dataset export — that's collection. Also "~12 style examples" overstates: the packs hold 9–10. The heading and the count are the two things still wrong here.

docs/lora-feasibility-spike.md
caiotheodoro: This paragraph is unchanged by the fix commit — the Qwen safetensors gap, GGUF conversion, single-adapter limit, and the step-0 adapter smoke test are still missing, and this is the load-bearing serving claim.

docs/lora-feasibility-spike.md
caiotheodoro: Still two variables in one A/B after the fix — the zero-training arm (base model + trimmed prompt) is what makes any voice_match delta attributable. Note #78 is now testing the cheaper curation lever directly, which strengthens the case for running that arm first.

matheusht:
@caiotheodoro ping — all four review findings are now addressed on this branch: weakest-axes claim scoped to the pinned baseline (only voice_match/in_character motivate the spike), cross-agent echo probe named accurately (it's on main now), seed-data accounting made honest (~12 style examples; 10–20 range only holds combined), and an explicit train/eval leakage exclusion added to both the experiment steps and known risks.

PR #77, opened 2026-08-22
feat(eval): jury-of-judges with per-axis median verdict

matheusht:
## What

Adds a **jury-of-judges** capability to the eval judge module:

- New `juryJudge(scenario, events, configs[])`: runs the same transcript through independently-sourced judge configs (different model family and/or endpoint per entry) and aggregates per-axis by **median** — the ordinal-appropriate majority verdict that measurably reduces same-family self-preference bias.
- Returns `perJudge` raw scores alongside the verdict: spread across judges IS the bias evidence this capability exists to detect. A single inflated outlier cannot move the median.
- Failure semantics: an errored judge is dropped (never averaged in); at least one survivor is required or the error is thrown honestly. Empty config list guarded.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+5 offline tests: median aggregation with per-judge inspection, outlier resistance, partial-failure survival, total-failure honesty, empty-config guard)
- Live cross-family comparison protocol documented in docs/eval/README.md (maintainer-run; uses the existing `PERFECTMAN_JUDGE_*` parameterization).


packages/eval/src/test/jury-judge.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          return Promise.resolve({ ok: false, status: 500, text: async () => "boom" });
+        }
+        return Promise.resolve({
+          ok: true,
+          status: 200,
+          json: async () => ({ choices: [{ message: { content: JSON.stringify({ axes: { in_character: 4 } }) } }] }),
+        });
+      }),
+    );
+    const verdict = await juryJudge(scenario, [], configs.slice(1));
```
caiotheodoro, line 133: bug: this partial-failure test is vacuous — the stub 500s only `family-a`, but the call passes `configs.slice(1)`, so the failing juror is never in the jury and the drops-failed-judges path runs zero times. Verified by mutation: passing the full `configs` (family-a in and failing) works with the identical assertions — the intended test is one edit away, the committed one tests nothing, and the PR body cites it as validating partial-failure survival.

packages/eval/src/judge/judge.ts
```diff
@@ -460,3 +460,88 @@ function hashString(s: string): string {
...
+      };
+    }
+  }
+  if (Object.keys(perJudge).length === 0) {
+    throw new Error("All jury judges failed");
+  }
+
+  // Salvaged scores are imputed defaults — they drag medians toward 3 and
+  // poison divergence evidence. Reported, but never voted.
+  const voters = Object.values(perJudge).filter(juror => !juror.salvaged);
```
caiotheodoro, line 533: bug: "imputed defaults never vote" is defeated one layer down — `parseAxes` fills every axis a juror omits with 3 and returns `salvaged:false`, so those unflagged 3s vote. Confirmed on head: a juror answering `{"axes":{"in_character":4}}` against the 8-axis rubric votes 3 on the other seven. The missing-axis filter at 543 is dead code for real llmJudge output, and with 2–3 voters one partial-JSON juror moves medians on axes it never scored. Track imputation per axis (e.g. `imputedAxes` on JudgeResult) and exclude those values like salvaged jurors.

packages/eval/src/judge/judge.ts
```diff
@@ -460,3 +460,88 @@ function hashString(s: string): string {
...
+
+  const axisIds = new Set<string>();
+  for (const juror of voters) {
+    for (const axis of Object.keys(juror.axes)) axisIds.add(axis);
+  }
+
+  const axes: AxisScores = {};
+  for (const axis of axisIds) {
+    const votes = voters.map(juror => juror.axes[axis]).filter((v): v is number => typeof v === "number");
+    if (votes.length > 0) axes[axis] = Math.round(median(votes) * 1000) / 1000;
```
caiotheodoro: bug: even-count juries emit half-integers (median [3,4] = 3.5, confirmed), and `weightedKappa` does `levels.indexOf(3.5)` → -1 → silently returns 0 for the axis (confirmed, no error). Since `JuryVerdict.axes` is the same type `calibrateJudge` consumes and the README steers headline numbers to the jury, the first jury calibration reads kappa=0 with no diagnostic — #24's exact failure signature. Note a 3-juror jury with one dropped voter quietly becomes an even jury. Round to the ordinal scale with a documented tie rule, or make weightedKappa throw on out-of-scale values.

packages/eval/src/judge/judge.ts
```diff
@@ -460,3 +460,88 @@ function hashString(s: string): string {
...
+ * mean and offers no outlier resistance. Salvaged jurors (prose-implied
+ * scores, mostly imputed defaults) are reported but EXCLUDED from medians.
+ * Judges that error are dropped; at least one unsalvaged survivor is
+ * required. Labels must be unique — duplicates would silently collapse
+ * two votes into one.
+ */
+export async function juryJudge(
+  scenario: RoleplayScenario,
+  events: readonly CommittedEvent[],
+  configs: Array<LLMJudgeConfig & { label?: string }>,
```
caiotheodoro, line 501: risk: family diversity — the entire point of #28/PoLL — is neither enforced nor recorded. Three qwen3:8b configs with distinct labels are a valid jury, and perJudge doesn't record model/baseUrl, so the evidence trail can't show the sourcing later. Throw (or warn) on duplicate (baseUrl, model), include each juror's model/baseUrl in perJudge, and state the disjoint-family requirement in the doc comment as a caller obligation.

packages/eval/src/judge/judge.ts
```diff
@@ -460,3 +460,88 @@ function hashString(s: string): string {
...
+  const settled = await Promise.allSettled(
+    configs.map(async (config, i) => ({
+      label: labels[i]!,
+      result: await llmJudge(scenario, events, config),
+    })),
+  );
+
+  const perJudge: Record<string, JuryJuror> = {};
+  for (const outcome of settled) {
+    if (outcome.status === "fulfilled") {
```
caiotheodoro, line 520: risk: failures degrade silently and discard their errors — a juror that 500s or times out (llmJudge retries only parse failures, not transport) vanishes with no trace except a missing perJudge key. Concretely: the one cross-family juror times out and the verdict quietly becomes a median of two same-family judges while looking like a healthy jury. Return a `failed: Record<label, string>` on JuryVerdict and consider an optional `minVoters`.

packages/eval/src/test/jury-judge.test.ts
```diff
@@ -0,0 +1,187 @@
...
+import { getScenario } from "@perfectman/shared";
+
+const scenario = getScenario("v1_casual_chat")!;
+
+const baseConfig = {
+  baseUrl: "http://judge-host/v1",
+  model: "test-model",
+};
+
+function stubJudgeResponses(responses: Array<Record<string, number>>): void {
```
caiotheodoro, line 12: nit: `stubJudgeResponses` is dead code after the body-routing migration — and it's exactly the order-dependent pattern the commit exists to eliminate. Delete it or make it the shared body-routing factory.

caiotheodoro (commented):
Three bugs and two risks inline, all verified by execution (8/8 committed tests pass, but see the first finding). The architecture is sound — allSettled degrade, salvage exclusion, label guard, and the body-routed stub refactor is the right determinism fix. PR-wide: juryJudge is exported but unreachable from bench/calibrate, while the README tells users to "switch to the jury for headline numbers" — a switch the tooling can't perform; a jury is also 3–6 judge calls per scene and nothing surfaces that cost; and jury-vs-single calibration semantics for #24 are unstated (see the kappa finding). One sentence each in the README would hold it until the follow-ups land.

matheusht:
@caiotheodoro ping — both medium findings fixed (duplicate labels throw before any fetch; salvaged jurors reported but excluded from medians, all-salvaged throws), ≥3-survivor regime documented, and test stubs are now body-routed so they're deterministic under concurrency.

PR #78, opened 2026-08-22
feat(shared): refresh persona style examples for creative range

matheusht:
## What

Refreshes persona-pack style examples so creative range isn't structurally capped:

- Each pack grows to 12 examples mixing signature anchors with new, more vivid in-voice lines: bruno's deadpan gets an "archive" motif ("não é rancor, é arquivo."), goulart gets theatrical self-awareness ("tô montando o próximo episódio"), leo gets chaotic specificity ("a melhor pior ideia que já ouvi na vida"), mariana's pure-filler lines are replaced with precise intrigue ("duas pessoas sabem disso. uma sou eu."), caio's mediation gains warmth and humor ("briga sem lanche não se resolve, fato").
- Three objective content tests pin the contract: ≥12 examples per pack, zero intra- or cross-pack duplicates, and a ≥2× short/long length spread so neither mock nor model anchors to one rhythm.
- Persona content only — prompt structure untouched (that work belongs elsewhere by agreement).

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+3)
- Canary-slice bench: 12 runs, signals 100%, BENCH GATE PASSED.
- Note: rule-judge `creativity_unhinged` cannot see text (emoji/leak-derived), so judged improvement is measurable only via live-model LLM-judge runs — maintainer follow-up.


caiotheodoro (commented):
Direction is right and the curation is genuine — ran the full 123-task mock bench on main vs this head: signals hold 100%, content-repetition mean drops 0.223→0.157, guard blocks −17% to −39%, no cross-agent echo. Two pack findings inline. The third blocker is evidence discipline: every mock benchmark derives from styleExamples, so this PR silently invalidates three committed artifacts (both bench-mock-baselines, the README's "Current baseline" numbers, and temperature-sweep-mock.json whose guardBlocks=92 reproduces on main but yields 76 here) — and rule-judge voice_match collapses 2.60→1.08 / in_character 3.57→2.93 purely because the new longer lines carry fewer STYLE_TELLS substrings (density 46%→34%): text better, proxy worse, mentioned nowhere. Commit a before/after report, refresh the baseline+sweep artifacts, and state the proxy shift with its mechanism — otherwise the next reader sees a regression with no explanation and the calibration alpha crash (−0.09→−0.80) reads as judge rot. The "creativity isn't measurable in mock" caveat in the body is accurate per the judge source.

packages/shared/src/persona-packs/caio.ts
```diff
@@ -21,15 +21,18 @@ export const CAIO_PACK: PersonaPack = {
     "Repair in private DMs first, public peacekeeping second.",
   ],
   styleExamples: [
+    "gente, respira comigo: tudo bem? tudo bem.",
```
caiotheodoro, line 24: bug: tell-cited exemplars deleted without replacement — the exact gap commit 85bdc82 fixed for Goulart, unfixed here. Caio's punctuationTells ("parenthetical honesty", :178) and maskTell ("'tô bem sim!!' gets louder as it gets less true", :153) lost their only exemplar; no remaining line has a parenthetical or "tô bem sim". Same for Bruno: "tudo bem sim :)" removed while bruno.ts:146/:153 and even leo.ts:55's memory seed still cite the phrase. Restore one in-voice exemplar each, e.g. Caio "tô bem sim!! (não tô, mas depois eu conto)".

packages/shared/src/persona-packs/bruno.ts
```diff
@@ -20,13 +20,16 @@ export const BRUNO_PACK: PersonaPack = {
     "Edit and re-edit messages before sending, often into something softer.",
   ],
   styleExamples: [
+    "ah, o bruno tem opinião agora? que novidade.",
```
caiotheodoro: risk: seven of the eight new Bruno lines are fully punctuated ("ri na minha cabeça. por dentro, óbvio.") against his own contract — voiceGuidelines[0] says "rarely use punctuation or capitalization", punctuationTells says "no punctuation" — and the terse two-beat aphorism shape mirrors Mariana, whose distinguishing tell IS perfect punctuation (mariana.ts:178). The cross-persona contrast blurs exactly where a model conditions on examples. Drop terminal periods from most lines ("não é rancor, é arquivo" already does it right) or amend his guideline in the same PR.

packages/shared/src/persona-packs/goulart.ts
```diff
@@ -20,15 +20,19 @@ export const GOULART_PACK: PersonaPack = {
     "Hide real feelings in parentheticals and quick denials ('tô nem aí (tô sim)').",
   ],
   styleExamples: [
+    "eu falo o que todo mundo pensa, alguém tinha que dizer",
+    "cadê a plateia? ah, é vocês",
+    "isso foi fraco até pro padrão daqui",
```
caiotheodoro: nit: "isso foi fraco até pro padrão daqui" (and Bruno's "vou fingir que não vi isso") hit the silence-misreading CONFRONT_LEXICON — verified in transcripts they now land within 2 events of silences and flipped motive_vulnerability out of band, the sole new probe failure (100%→97%). Mostly a mock artifact, but these are also the lines a live model imitates after someone goes quiet.

matheusht:
@caiotheodoro ping — goulart's parenthetical-honesty exemplar restored (guideline cites it) and the stray blank-line churn removed. 16/16 pack tests green.

Issue #32, opened 2026-08-20
unpredictability is flat at 3.0 across both benchmark rounds


caiotheodoro:
`unpredictability` is the one edge_chaos-shared axis that showed zero movement across this session's before/after benchmark rounds — 3.0 in both. Every other axis this session touched (directly or incidentally) moved at least somewhat; this one didn't move at all, suggesting none of the fixes this session made (validation guards, repetition retry, memory scaffolding) have any real relationship to how unpredictable the agents' choices feel.

Needs a dedicated hypothesis and targeted intervention, not incidental improvement from fixing something else. Possible directions: sampling-parameter experiments (temperature/top_p tuning specifically for this axis), or prompt-level encouragement toward less-expected-but-still-in-character choices.

**Where to look**: `packages/eval/src/judge/rubrics.ts`'s `EDGE_CHAOS_RUBRIC` (the anchor descriptions for this axis), `packages/server/src/agent/persona-loader.ts` (sampling config).

matheusht:
## Root cause

The rule-judge proxy computes `unpredictability = 3 + emojiRate×2`. With zero reactions in a room it is structurally pinned to exactly 3.0 — which is precisely what both benchmark rounds reported. The metric was measuring nothing about choice diversity.

## Fix

Replace the proxy with **normalized intent-type entropy**: treat the scene's committed (non-operator) event types as the observable choice distribution, compute Shannon entropy normalized by log(distinct types), and map to the axis scale (`1 + H·4`, range 1–5):

- one action type everywhere → 1.0 ("fully predictable boilerplate"),
- k types spread uniformly → ~5 ("chaotic in the best way").

This measures the anchor text directly ("unpredictable CHOICES"), works offline deterministically, and finally lets sweeps/CI gates track whether interventions change behavioral diversity. LLM-judged unpredictability is untouched.

Tests: single-type floor, uniform-ceiling, monotonicity across diversities, operator-event exclusion.

Closes #32 once shipped.

yoarajota:
Addressed by PR #79: the intent-entropy proxy (per-agent Shannon entropy over choice types, log-5 denominator, channel-batch dedupe) replaced the `3 + emoji*2` proxy that was structurally pinned at 3.0 in any room that never reacts. The axis now differentiates (gate mean measured 4->3, uniform-5 room scores 5). Tests in eval/src/test/intent-entropy.test.ts.

PR #79, opened 2026-08-22
feat(eval): intent-entropy proxy for the unpredictability axis

matheusht:
## What

Replaces the rule-judge `unpredictability` proxy with a metric that can actually move:

- **Old proxy**: `3 + emojiRate×2` — structurally pinned at exactly 3.0 whenever a room commits no reactions, which is why the axis reported zero movement across benchmark rounds while every other axis moved. It measured reactions, not unpredictability.
- **New proxy**: normalized Shannon entropy over the distribution of committed choice types (message/reply/react/no-op/channel ops; operator and spectator events excluded). Maps to the axis scale via `1 + H·4`: one dominant action type scores the floor, uniformly spread choices approach 5 — matching the anchor text ("unpredictable CHOICES within persona" → "never the expected move").
- Deterministic, offline, dependency-free; LLM-judged runs are untouched.
- Five unit tests: single-type floor, uniform ceiling, skew monotonicity, operator-noise exclusion, empty input.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+5)
- Head-to-head on edge scenarios: `edge_mutation_pressure` moves **3 → 4** (its genuine choice diversity finally registers); `edge_public_mock` holds at 4.


packages/eval/src/judge/judge.ts
```diff
@@ -153,6 +153,47 @@ export function ruleJudge(
...
+    counts.set(e.type, (counts.get(e.type) ?? 0) + 1);
+  }
+  const total = [...counts.values()].reduce((s, c) => s + c, 0);
+  if (total === 0 || counts.size <= 1) return 0;
+  let entropy = 0;
+  for (const count of counts.values()) {
+    const p = count / total;
+    entropy -= p * Math.log(p);
+  }
+  return entropy / Math.log(counts.size);
```
caiotheodoro: bug: normalizing by `Math.log(counts.size)` measures evenness of whatever appeared, not repertoire breadth — a room that only messages and replies 50/50 (the most ordinary chat shape) scores H=1 → unpredictability 5 at any length. Measured: all four edge scenarios land in H 0.66–0.76, which `clampScore` rounds to a constant 4 — flat at 4 instead of flat at 3. Normalize by `Math.log(CHOICE_EVENT_TYPES.size)` (log 6) instead: the 2-type room drops to ~3, the uniform-6-type room stays 5, and the tiny-transcript ceiling fixes itself. The uniform-ceiling test at intent-entropy.test.ts:31 currently locks in the wrong denominator and needs updating with it.

caiotheodoro (commented):
One bug and one risk inline, both measured with the repo's own calibrate CLI on main vs this head. The implementation is clean and the tests pass (90/90), but the headline claim doesn't hold: under the calibration harness the axis goes from near-constant (4,4,4,3) to fully constant 4, per-axis kappa stays exactly 0.000, overall kappa 0.129 → 0.129, alpha slightly worse — and the celebrated edge_mutation_pressure 3→4 is movement AWAY from its golden label (all four edge goldens say 3). Per the repo's own committed-never-hidden doctrine, this judge change needs the recomputed calibration report committed, and either a golden-label position or an honest note that the flat-3.0 in #32 came from real rounds, not the mock harness. Also note bench-mock-baseline.json goes stale for this axis.

packages/eval/src/judge/judge.ts
```diff
@@ -153,6 +153,47 @@ export function ruleJudge(
...
+
+/**
+ * Normalized Shannon entropy over the distribution of committed choice
+ * types: 1.0 when the room's actions are spread uniformly across its
+ * distinct types, 0.0 when every action is the same type. This replaces
+ * the old `3 + emoji*2` proxy, which was structurally pinned at 3.0 in any
+ * room that never reacts (#32) — the anchor measures unpredictable
+ * CHOICES, so the proxy should measure choice diversity.
+ */
+export function intentEntropyScore(events: readonly CommittedEvent[]): number {
```
caiotheodoro, line 211: risk: entropy is pooled across the room but the rubric anchor is per-agent ("unpredictable choices within persona") — agent A always messaging plus agent B always reacting pools to high entropy while both are perfectly predictable. Mock agents behave alike so pooled ≈ per-agent today (0.72–0.80 vs 0.76 on edge_public_mock); real LLM rooms are exactly where they diverge. Also one create-channel choice emits channel_created + N agent_invited events, inflating two buckets from a single decision. Compute per actorId and average (or min), and count the channel batch as one choice.

packages/eval/src/judge/judge.ts
```diff
@@ -153,6 +153,47 @@ export function ruleJudge(
...
+]);
+
+/**
+ * Normalized Shannon entropy over the distribution of committed choice
+ * types: 1.0 when the room's actions are spread uniformly across its
+ * distinct types, 0.0 when every action is the same type. This replaces
+ * the old `3 + emoji*2` proxy, which was structurally pinned at 3.0 in any
+ * room that never reacts (#32) — the anchor measures unpredictable
+ * CHOICES, so the proxy should measure choice diversity.
+ */
```
caiotheodoro, line 210: nit: content-blindness isn't stated — an agent that always sends messages, each wildly surprising in content, scores 0 → axis floor 1 (previously 3). One docstring sentence: type-level only, surprising content within a fixed type is invisible to this proxy; the LLM judge owns that. Same pass: move the tiny-transcript caveat above the `return` — after it is the one place nobody reads.

packages/eval/src/judge/judge.ts
```diff
@@ -153,6 +153,47 @@ export function ruleJudge(
...
+export function intentEntropyScore(events: readonly CommittedEvent[]): number {
+  const counts = new Map<string, number>();
+  for (const e of events) {
+    if (!CHOICE_EVENT_TYPES.has(e.type)) continue;
+    counts.set(e.type, (counts.get(e.type) ?? 0) + 1);
+  }
+  const total = [...counts.values()].reduce((s, c) => s + c, 0);
+  if (total === 0 || counts.size <= 1) return 0;
+  let entropy = 0;
+  for (const count of counts.values()) {
```
caiotheodoro: nit: degenerate rooms now land at the floor — all-no_op and empty transcripts score 1 where they scored 3. Defensible per anchor 1, but it shifts the axis mean for silent rooms; worth a line in the PR body.

matheusht:
@caiotheodoro ping — tiny-transcript ceiling caveat documented in the proxy JSDoc (2 events/2 types score the ceiling — fine for a v0 proxy). Verified head-to-head: edge_mutation_pressure unpredictability moves 3→4 with genuine choice diversity.

PR #86, opened 2026-08-23
feat(judge): judge as a config-file section — shared schema, server loader, eval precedence

yoarajota:
## Judge as a config-file section — PIVOT (for review)

Main is the exact pre-pivot state (`364859a` + revert commits `3b0c1f3`/`2c16f39`); this PR is the structure half of the judge-config pivot:

- **Shared schema** (`packages/shared/src/judge/`): `JudgeAppConfig`, types **z.infer-derived** from the schemas (the intent-packet convention) with a type-level drift test. `providerType: "openai-compatible" | "rule" | "mock"`, `baseUrl`, `modelName`, `apiKeyEnv` (secret by NAME, agent pattern), `temperature`, `timeoutMs`, `retryCount`, `responseFormatJson`, plus the `jury` array. DeepSeek is no longer a provider type. Parse-time refinements mirror the runtime jury guards: duplicate labels (including judge-N default collisions) and duplicate (baseUrl, modelName) endpoints are rejected before any judge runs; inline secrets rejected like the agent LLM section.
- **Server loader** (`config/judge-config.ts`): `loadJudgeConfig(path)` + `loadJudgeSectionFromSimulationConfig` — the walk-up for the `judge` section now **stops at the workspace root** (`pnpm-workspace.yaml`), so a stray `~/config/index.json` cannot repoint or hard-fail runs — + `getJudgeConfigPath` in `config-path.ts` (generalized resolver, `getConfigPath` semantics unchanged). `SimulationAppConfig` gains `judge?`.
- **Eval loader** (`llm/judge-config.ts`): precedence `--judge-config <path>` > `config/index.json` `judge` section > env > defaults. The env tier is byte-identical to the pre-pivot `judgeConfig()` (verified across 8 env profiles, re-verified after the review fixes). `judgeConfig()` survives as a thin **envOnly** compat call (no file tiers). File-tier secrets resolve **only** via `apiKeyEnv` — the ambient `PERFECTMAN_LLM_API_KEY` never rides along to a file-named endpoint. Validated-but-unconsumed keys (`responseFormatJson`; `retryCount` until the stacked superset PR) and unknown keys (typo surface, e.g. `baseURL`) **warn at load** instead of silently no-op'ing.
- **Provider dispatch**: `--judge rule|llm` is an explicit shorthand override of `providerType`; `rule`/`mock` (deterministic `MockJudgeProvider`)/`openai-compatible` dispatch; the file's `jury` resolves through the same loader. `bench --judge llm` with a jury runs the median verdict and threads **`juryVoterCount`/`juryFailed` into the report** (a 3-juror jury with 2 errors is a traceable single-judge read, per #77/#81 visibility); `--per-turn` + jury now **throws** instead of silently ignoring.
- **Docs**: `docs/eval/README.md` gains the file-form example, reserved-keys limits, and the calibrate/jury note.

**Known deltas to flag (reviewer-requested):**
- `calibrate` now honors `PERFECTMAN_JUDGE_BASE_URL` where pre-pivot it consulted only `PERFECTMAN_LLM_BASE_URL` — almost certainly an improvement; the "env tier byte-identical" claim covers **bench** (and the compat `judgeConfig()`), with calibrate inheriting the unified loader.
- If this PR merges without the stacked superset (#87), `retryCount` and `responseFormatJson` are dead config keys — the loader warns about both.

Validation: build/typecheck 4/4; **97 files / 1014 tests**; audit 922 its, 0 violations; gate bench **35 scenarios, signal 100.0%, probe pass 89.9% — PASSED**; both sweep evidence files byte-identical; env-only judge resolution byte-identical across 8 profiles.


packages/eval/src/cli/bench.ts
```diff
@@ -176,11 +167,18 @@ export async function runBench(opts: {
...
       const judgeResult =
-        judgeMode === "llm"
-          ? perTurn
-            ? await llmJudgePerTurn(scenario, artifact.events, judgeConfig())
-            : await llmJudge(scenario, artifact.events, judgeConfig())
-          : { axes: ruleJudge(scenario, artifact.events, artifact.probeResults, artifact.passedSignals / Math.max(1, artifact.totalSignals)), salvaged: false };
+        judgeMode === "openai-compatible"
+          ? resolvedJudge.jury && resolvedJudge.jury.length > 0
+            ? {
+                axes: (await juryJudge(scenario, artifact.events, resolvedJudge.jury)).axes,
```
caiotheodoro: bug: the jury path keeps only `.axes` and hardcodes `salvaged: false`, discarding `failed`, `perJudge`, and `voterCount` — a 3-juror jury with 2 errored jurors silently degrades to a single-judge verdict with zero trace, which is exactly the failed-juror visibility #77/#81 established (and the median's outlier resistance holds "ONLY when >= 3 jurors survive" per juryJudge's own doc). Thread `voterCount` and `failed` into the report, or at minimum warn when `failed` is non-empty.

packages/shared/src/judge/judge-config.schema.ts
```diff
@@ -0,0 +1,69 @@
...
+  .superRefine(rejectInlineSecrets);
+
+export const JudgeAppConfigSchema = judgeConfigObject
+  .extend({ jury: z.array(JudgeEntryConfigSchema).optional() })
+  .superRefine(rejectInlineSecrets)
+  .superRefine((config, ctx) => {
+    // Mirror of the runtime guard in eval's juryJudge (`Duplicate jury
+    // judge labels`) — the file is rejected before any judge ever runs.
+    const labels = (config.jury ?? []).map((j) => j.label ?? "");
+    const duplicates = labels.filter((label, i) => label !== "" && labels.indexOf(label) !== i);
```
caiotheodoro: risk: the parse-time guard is half the runtime one — unlabeled entries are skipped here but the runtime defaults labels to `judge-${i}`, so a partially-labeled jury passes parse and throws mid-bench; and the runtime endpoint-disjointness check ("a jury must be independently sourced") has no parse-time mirror, so a same-endpoint jury dies only after the first scenario's generation completes. Apply both in the superRefine, and note in the doc that disjointness is endpoint-level — family diversity (the #77 requirement) still isn't expressible.

packages/eval/src/cli/bench.ts
```diff
@@ -176,11 +167,18 @@ export async function runBench(opts: {
...
+                salvaged: false,
+              }
+            : perTurn
```
caiotheodoro: risk: `--per-turn` is silently ignored when the file configures a jury — the jury branch short-circuits before the perTurn ternary. Throw on the combination or route each juror through `llmJudgePerTurn`.

packages/eval/src/cli/bench.ts
```diff
@@ -20,37 +20,22 @@ import {
...
-    // Set PERFECTMAN_JUDGE_TEMPERATURE=0 for deterministic calibration runs.
-    temperature: tempSet && Number.isFinite(tempRaw) ? tempRaw : 1.0,
-    timeoutMs: 90000,
-  };
+/**
+ * Compatibility view of the judge loader for programmatic callers: the
+ * env-resolved endpoint config, exactly as the pre-pivot judgeConfig()
+ * returned it (no file, no CLI shorthand).
+ */
+export function judgeConfig(): Promise<import("../judge/judge.js").LLMJudgeConfig> {
```
caiotheodoro, line 38: risk: the compat `judgeConfig()` is now async (breaking for programmatic callers) and its "env-resolved, no file" docstring is wrong — `loadJudgeConfig([])` still walks up to a `config/index.json` judge section, and the walk-up continues past the repo root, so a stray `~/config/index.json` can repoint or hard-fail every bench/calibrate run including rule-judge gate runs. Add an envOnly option for the compat path and stop the walk-up at the workspace root.

packages/eval/src/llm/judge-config.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+function resolveEntry(
+  entry: JudgeConfigBase,
+  envDefaults: EnvJudgeDefaults,
+): LLMJudgeConfig {
+  return {
+    baseUrl: entry.baseUrl ?? envDefaults.baseUrl,
+    model: entry.modelName,
+    apiKey: entry.apiKeyEnv ? process.env[entry.apiKeyEnv] : envDefaults.apiKey,
```
caiotheodoro: risk: file-tier entries without `apiKeyEnv` silently inherit `PERFECTMAN_LLM_API_KEY` as the bearer for whatever baseUrl the file names — the ambient secret can be sent to an unrelated endpoint listed in a shared config file, contradicting the resolve-by-name contract. In the file tier, resolve keys only via `apiKeyEnv`, or restrict the fallback to the default baseUrl. (The rest of the secret hygiene checks out: name-only resolution, inline keys rejected, no config in reports.)

packages/eval/src/llm/judge-config.ts
```diff
@@ -0,0 +1,135 @@
...
+    temperature: tempSet && Number.isFinite(tempRaw) ? tempRaw : defaultTemperature,
+    timeoutMs: 90000,
+  };
+}
+
+function resolveEntry(
+  entry: JudgeConfigBase,
+  envDefaults: EnvJudgeDefaults,
+): LLMJudgeConfig {
+  return {
```
caiotheodoro, line 78: risk: `resolveEntry` drops `responseFormatJson` (and `retryCount` until #87) after successful parse — silent no-ops — and `.passthrough()` means typo'd keys like `baseURL` are silently ignored while the env fallback quietly wins. Also no `extraBody` in judge config, so the default local judge can't disable think-mode by config and survives only via the parser's think-stripping. Document the reserved keys in the README example and warn on unconsumed ones.

packages/eval/src/cli/calibrate.ts
```diff
@@ -94,7 +87,7 @@ export async function main(): Promise<void> {
     // scored the transcripts. Golden labels were authored against the mock
     // baseline, so LLM-judged calibration reads mock transcripts too.
     mode: "mock" as const,
-    judge: judgeMode,
+    judge: providerType === "openai-compatible" ? "llm" : "rule",
```
caiotheodoro: nit: a `providerType: "mock"` run is recorded as `judge: "rule"` in the calibration report — evidence mislabeling. Also: a configured jury is ignored by calibrate (single llmJudge), unlike bench — worth one sentence somewhere.

packages/eval/src/cli/calibrate.ts
```diff
@@ -24,31 +26,20 @@ function argValue(flag: string): string | undefined {
...
+  const judgeFlag = argValue("--judge");
+  if (judgeFlag !== undefined && judgeFlag !== "rule" && judgeFlag !== "llm") {
+    console.error(`unknown --judge value: ${judgeFlag} (expected rule|llm)`);
     process.exit(2);
   }
   const out = argValue("--out");
 
+  // Calibration is an agreement measurement — determinism by default
+  // (temperature 0 rather than bench's heuristic 1.0).
+  const resolvedJudge = await loadJudgeConfig(args, { defaultTemperature: 0 });
```
caiotheodoro, line 39: nit: calibrate now honors `PERFECTMAN_JUDGE_BASE_URL` where pre-pivot it consulted only `PERFECTMAN_LLM_BASE_URL` — almost certainly an improvement, but the "env tier byte-identical" claim covers bench only; flag this calibrate delta in the body.

packages/shared/src/judge/judge-config.schema.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (duplicates.length > 0) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: `Duplicate jury judge labels: ${[...new Set(duplicates)].join(", ")}`,
+      });
+    }
+  });
```
caiotheodoro: nit: five of the new files end without a trailing newline.

caiotheodoro (commented):
One bug and five risks inline; the precedence/back-compat engineering is well done. Verified: the re-raise is byte-identical to the merged-then-reverted #83 (identical tree hashes), nothing landed in docs/eval/evidence during the 24-minute merged window, and the full gate passes on this head (1001/1001 tests, signals 100%). Two PR-wide notes: judge-config.types.ts is hand-written rather than z.infer-derived and has no drift test — the packet-schema convention this claims to follow; and if #86 merges without #87, both retryCount and responseFormatJson are dead config keys.

PR #87, opened 2026-08-23
feat(llm): chatCompletion promoted to the shared superset; provider adopts it

yoarajota:
## Shared chat-completion superset — re-raised for review (stacked on #86)

Same content lineage as the previously merged-then-reverted #84, re-raised
so this refactor lands only after maintainer review.

- **Superset primitive** (`packages/shared/src/llm/chat-completion.ts`,
  moved from `packages/eval/src/llm/`): transport retry (`retryCount`,
  transient = timeout / 429 / 5xx / network), **dual timeout** (connect
  AbortController + body-read and error-body-read deadlines raced against
  hard timers that are **cleared on every settle** — no dangling timer
  holds eval CLIs open after the last LLM call), **response-header
  extraction** (lowercased), **typed errors** aligned with the server
  `LLMError` hierarchy (`ChatCompletionHttpError` status+body,
  `ChatCompletionTimeoutError`). json_schema response format with the
  automatic 400/422 → json_object fallback on the same budget slot,
  surfaced as `ChatCompletionResult.schemaFallbackUsed` (the #51-flagged
  degrade is no longer invisible). Body precedence is the provider's:
  named knobs first, then `extraBody` wins, then response_format —
  pinned by tests.
- **`OpenAiCompatibleProvider.generateIntent` adopts the primitive** —
  the transport loop becomes the shared call path; the provider keeps
  packet JSON Schema selection, usage/route accounting, and the LLMError
  boundary wrapping (message parity preserved, incl. `fallbackAttempts`
  net-attempt arithmetic).
- **Judge + narrator follow** from `@perfectman/shared`;
  `LLMJudgeConfig.retryCount` additive — the file's `retryCount` now
  drives transport retries on judge calls (threaded in the eval loader;
  consequently removed from the unconsumed-key warning).
- `@types/node` added to `@perfectman/shared` devDeps. **Lockfile note:
  the delta also re-resolves `@types/ws`'s node types — harmless, but the
  body previously oversold it as "exactly" the one addition.**

Validation: provider suites **31/31** (all 9 adopted-path pins:
result structure/routedModel/fallbackAttempts, config errors, HTTPError,
TimeoutError, json_schema body, json_object force, 400 fallback 2 calls
with net attemptCount 1, malformed-shape fail-fast 1 call, 429 retries
3 calls); **97 files / 1027 tests**, audit 935 its, 0 violations; gate
bench **35 scenarios, signal 100.0%, probe pass 89.9% — PASSED**; both
sweep evidence files byte-identical; env-only judge resolution
byte-identical across 8 profiles.

Merge order: **#86 first** (hard dependency — `retryCount` lives in its
schema).


packages/shared/src/llm/chat-completion.ts
```diff
@@ -0,0 +1,233 @@
...
+          response.status,
+          `${opts.label} HTTP ${response.status}: ${errorText}`,
+          errorText,
+        );
+      }
+
+      // The abort timer stays armed through the body read — response.json()
+      // can hang on a slow stream after headers arrive, so race the body
+      // read against a hard deadline too.
+      const deadline = new Promise<never>((_, reject) =>
```
caiotheodoro: bug: the body-read deadline timer is a bare setTimeout inside the Promise.race and is never cleared — only `timeoutId` is (198). Reproduced against the built package: a call resolving in 0ms held process exit 3002ms with timeoutMs 3000. The deleted eval primitive used AbortSignal.timeout, whose timer is unref'd, so every eval CLI (narrate, calibrate, sweep-*) now sits idle up to 60–90s after its last LLM call before exiting. Capture the deadline timer id and clear it in a finally alongside timeoutId.

packages/shared/src/llm/chat-completion.ts
```diff
@@ -0,0 +1,233 @@
...
+  const timeoutMs = opts.timeoutMs ?? 60000;
+  const url = `${opts.baseUrl.replace(/\/$/, "")}/chat/completions`;
+
+  const body: Record<string, unknown> = {
+    model: opts.model,
+    messages: opts.messages,
+    max_tokens: opts.maxTokens,
+    ...opts.extraBody,
+  };
+  if (opts.temperature !== undefined) body["temperature"] = opts.temperature;
```
caiotheodoro, line 161: risk: behavior change hidden as refactor — temperature is assigned AFTER the `...opts.extraBody` spread, but the old provider spread extraBody last, so a configured `extraBody.temperature` override now silently stops working (while max_tokens keeps the old extraBody-wins order — the two knobs are now inconsistent). No test pins the precedence in either suite. Restore extraBody-wins, or document named-param-wins as the new contract and pin it.

packages/shared/src/llm/chat-completion.ts
```diff
@@ -0,0 +1,233 @@
...
+  if (opts.responseFormatJson) {
```
caiotheodoro: risk: `jsonSchemaFormat` is only consulted inside `if (opts.responseFormatJson)`, but its doc says "takes precedence when set" — a caller passing the schema without the flag silently gets no response_format at all. Today's one caller passes both, but this is now shared public API and the doc misleads the next one. Gate on either flag.

packages/shared/src/llm/chat-completion.ts
```diff
@@ -0,0 +1,233 @@
...
+      const response = await fetch(url, {
+        method: "POST",
+        headers,
+        body: JSON.stringify(body),
+        signal: controller.signal,
+      });
+
+      if (!response.ok) {
+        clearTimeout(timeoutId);
+        const errorText = await response.text().catch(() => "Unknown error");
```
caiotheodoro: nit: on non-2xx, timeoutId is cleared before `await response.text()` and the error-body read races no deadline — a proxy hanging mid-error-body hangs the call unboundedly. It's the one window with zero timeout coverage; race it against the same deadline pattern.

packages/shared/src/llm/chat-completion.ts
```diff
@@ -0,0 +1,233 @@
...
+
+      if (isTransient(resolvedError) && attempts < maxAttempts) {
+        continue;
+      }
+      throw resolvedError;
+    }
+  }
+
+  throw new ChatCompletionError("Max retries exceeded without resolving error.");
+}
```
caiotheodoro: nit: no trailing newline at EOF (openai-compatible-provider.ts too). Matches merged #84 exactly, so fixing it breaks the byte-identity claim — which is fine now.

caiotheodoro (commented):
One reproduced bug and two risks inline, all in chat-completion.ts. Verified: the delta over #86 is byte-identical to the merged-then-reverted #84 (empty diff against 533a663), OllamaProvider is untouched by the whole stack so #51's format degrade, think handling, and seed nesting are intact by non-touch, fallbackAttempts arithmetic is preserved exactly, and everything is green (1010/1010 tests, gate PASS, full 123-scenario bench matches the body to the decimal). Fixing the findings breaks the byte-identity provenance claim — fine, this review has independently established it. Two notes: the #51-flagged degrade invisibility (the 400/422→json_object extra call appears nowhere in attemptCount) is now codified into the shared contract — this promotion was the natural moment for a fallbackUsed flag on ChatCompletionResult; intentional deferral? And the lockfile isn't "exactly" the @types/node add — it also re-resolves @types/ws's node types (harmless). Merge #86 first (hard dependency: retryCount lives in its schema).

caiotheodoro (changes requested):

PR #88, opened 2026-08-23
feat(agent): repetition policy knobs ported onto the surface refactor + sweep

matheusht:
## What

Ports configurable repetition-guard policy onto the post-refactor architecture and closes the measurement gaps from the first iteration:

- **Policy store**: `setRepetitionPolicy({ threshold, maxRetries })` in repetition-guard.ts — a guarded module-level singleton mirroring `llmBudget`, clamped on write (threshold 0–1, integer retries ≥ 0). Defaults reproduce shipped behavior exactly.
- **ActionIntentStep integration**: threshold reaches `isNearRepeat`; retries run as a bounded loop with a **budget re-check before each wire call** (a retry budget can no longer bypass the token gate); correction prompts escalate by attempt number; the block motive states the actual attempt count (`maxRetries=0` → "without any retry attempt").
- **Sweep CLI** (`sweep:repetition`): thresholds × retry budgets over the canary set, seed-pinned. Per-cell metrics now include **wireCalls** instrumented at the provider layer — turn-level counting was structurally blind to retry cost. The content-repetition yardstick stays fixed at 0.7 and is recorded per cell as `probeThreshold`.
- **Evidence honesty**: the committed mock matrix embeds a `limitations` field stating threshold cells are indistinguishable by construction (~1.0 Jaccard repeats) — only wire-calls vary. Discriminating thresholds requires real-model runs (maintainer follow-up).
- Four knob tests: two-retry recovery, zero-budget immediate block, strict-threshold pass-through at Jaccard 5/6, default-threshold block.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+4 knob tests; existing repetition suite updated for attempt-accurate wording)
- Sweep executed end-to-end offline; wire-calls scale 398→520→642 across retry budgets as expected.


packages/server/src/agent/repetition-guard.ts
```diff
@@ -40,6 +40,34 @@ export function similarity(a: string, b: string): number {
...
+ * singleton; sweeps and tests set it per run via setRepetitionPolicy().
+ */
+export type RepetitionPolicy = { threshold: number; maxRetries: number };
+
+let activePolicy: RepetitionPolicy = {
+  threshold: REPETITION_SIMILARITY_THRESHOLD,
+  maxRetries: 1,
+};
+
+export function setRepetitionPolicy(policy: { threshold?: number; maxRetries?: number }): void {
```
caiotheodoro, line 58: risk: process-global `activePolicy` means two concurrent eval runs clobber each other. Main already threads this per run (`ScenarioRunner.run({ repetition })` → `ctx.repetitionPolicy`). Don't reintroduce the singleton.

caiotheodoro (changes requested):
Don't merge: #69 already landed the knobs/sweep, then main kept going (per-run `repetition`, empty-set Jaccard=0 in #80, schema decoding in the SDK). One inline. Close it; the only leftover worth a follow-up is the attempt>1 escalating retry note.

packages/server/src/agent/surface/action-intent-step.ts
```diff
@@ -130,50 +134,90 @@ export class ActionIntentStep implements LLMStep<AgentRuntimeInput, AgentRuntime
...
     const mainOutputTokens = providerResult.usage.outputTokens;
-    let retryUsage: { inputTokens: number; outputTokens: number; latencyMs: number; promptVersion: string } | undefined;
+    const retryUsages: Array<{ inputTokens: number; outputTokens: number; latencyMs: number; promptVersion: string }> = [];
 
+    const { threshold, maxRetries } = getRepetitionPolicy();
     const isRepeat = (candidateIntent: typeof intent): boolean =>
       (candidateIntent.intentType === "send_message" || candidateIntent.intentType === "reply_to_message") &&
       !!candidateIntent.visibleContent &&
-      isNearRepeat(candidateIntent.visibleContent, input.perceptionPacket.ownRecentUtterances);
+      isNearRepeat(candidateIntent.visibleContent, input.perceptionPacket.ownRecentUtterances, threshold);
```
caiotheodoro, line 144: bug: `similarity` still scores two empty-normalized strings as 1.0, so `isNearRepeat("😂", ["🔥"])` is a false positive regardless of this threshold. Return 0 from that branch — main already does this in #80.

caiotheodoro:
Closing: #69 already landed the knobs/sweep, and main kept going (per-run `repetition`, empty-set Jaccard in #80, SDK schema decoding). Don't rebase this. If the attempt>1 escalating retry note is still wanted, that's a small follow-up on main.

PR #99, opened 2026-08-26
fix(shared): drop dead operator event types; guard producers

Elesiann:
## What

`rate_limit_hit` is declared in the shared operator-event contract but never emitted anywhere in `packages/server/src` — `RateLimitGate` computes block state (`blocked`/`blockReason`) but never produces an operator event for it. A declared operator type with no producer can never reach a receiver; keeping it invites drift between the contract and the runtime. It was the only dead member of the union: every other declared type (`agent_state_snapshot`, `action_intent`, `event_visibility` included) has at least one emission site on main.

- **Removes `rate_limit_hit`** (`packages/shared/src/operator/operator.types.ts`) and turns the union into a runtime-iterable `as const` array with the type derived from it (`OPERATOR_EVENT_TYPES` = single source of truth; no zod schema added, per the testing-strategy won't-do).
- **Adds a producer guard** (`packages/server/src/simulation/__tests__/operator-event-producers.test.ts`): every declared operator type must have an emission site (`type: "<name>"`) in `packages/server/src`. The reverse direction — emitting an undeclared type — is already rejected by the type system, so the guard is one-directional. It validates all ten remaining types, including the observability-stream ones that landed on main via #100.
- Regenerates `docs/test-inventory.{md,json}`.

## Validation

- `pnpm lint` PASS (typecheck, 4/4 packages)
- `pnpm test:all` PASS — full suite green (100 files / 1064 tests incl. the new guard test); audit gate 0 violations
- The guard test passes against the 10 produced types and fails for any declared type without an emission site — the invariant this PR exists to enforce
- Gate run inside a `node:22` container (Docker, on the workspace's remote worker) with `pnpm install --frozen-lockfile`, matching the CI sequence (install → build → lint → test:all)

### DIFF
--- a/packages/server/src/simulation/__tests__/operator-event-producers.test.ts
+++ b/packages/server/src/simulation/__tests__/operator-event-producers.test.ts
@@ -18,8 +18,8 @@ import { OPERATOR_EVENT_TYPES } from "@perfectman/shared";
  * the type system already rejects any `type: "..."` that is not a member of
  * the union.
  *
- * This guard exists because `rate_limit_hit` was declared but never
- * emitted anywhere in packages/server/src — removed with this change. The
+ * This guard exists because `rate_limit_hit` was declared but never emitted
+ * anywhere in packages/server/src — removed with this change. The
  * observability-stream types (`agent_state_snapshot`, `action_intent`,
  * `event_visibility`) landed on main with PR #100 and each has a producer.
  */
@@ -44,6 +44,37 @@ function collectSourceFiles(dir: string): string[] {
   return files;
 }
 
+/**
+ * True if `type` is emitted as an OperatorEvent somewhere in `source`.
+ *
+ * A bare `type: "<name>"` occurrence is not proof of an operator emission:
+ * `intent_blocked`, `intent_delayed` and `llm_failure` are also committed
+ * `SimulationEvent` types, and the same `type:` literal appears in their
+ * constructions (`intent-resolver.ts` builds committed events with
+ * `type: "intent_blocked"`/`"intent_delayed"`; `pulse-scheduler.ts` builds a
+ * committed `llm_failure`). `detail` is a required field of `OperatorEvent`
+ * and never appears on `SimulationEvent`, so an occurrence with a `detail`
+ * property nearby is structurally an operator emission; one without (a
+ * committed-event construction, or a stray mention) must not count.
+ *
+ * Residual limitation: a doc comment that happens to mention both the type
+ * literal and `detail` within the window would count — none exists today.
+ */
+function hasOperatorEmission(source: string, type: string): boolean {
+  const needle = `type: "${type}"`;
+  let from = 0;
+  for (;;) {
+    const at = source.indexOf(needle, from);
+    if (at === -1) return false;
+    // `detail` is the operator-only discriminator; match both the explicit
+    // `detail:` form and the shorthand `detail,` (pulse-scheduler.ts uses
+    // shorthand in its schedulerError helper).
+    const window = source.slice(Math.max(0, at - 200), at + 300);
+    if (/detail\s*[:,]/.test(window)) return true;
+    from = at + needle.length;
+  }
+}
+
 describe("operator event producers", () => {
   const sourceFiles = collectSourceFiles(SERVER_SRC);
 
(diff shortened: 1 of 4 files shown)


packages/shared/src/operator/operator.types.ts
```diff
@@ -1,15 +1,27 @@
...
+/**
+ * Operator event types the server may emit. Declared once here — the single
+ * source of truth — and enforced by the producer guard in packages/server
+ * (`operator-event-producers.test.ts`): a type with no emission site in
+ * packages/server/src is dead and must be removed, not kept "for later".
+ * Deriving the type from the array keeps the runtime-iterable list and the
+ * type in one place.
+ *
+ * `rate_limit_hit` was removed (declared, never emitted).
+ * `agent_state_snapshot` was removed (declared, never emitted on main).
```
caiotheodoro: bug: `agent_state_snapshot` is emitted on current main from `persistAndSnapshot` in `pulse-scheduler.ts`. Keep it.

packages/shared/src/operator/operator.types.ts
```diff
@@ -1,15 +1,27 @@
...
+ */
+export const OPERATOR_EVENT_TYPES = [
+  "llm_failure",
+  "llm_budget_exceeded",
+  "intent_blocked",
+  "intent_delayed",
+  "stagnation_warning",
+  "scheduler_error",
+  "pulse_metrics",
+] as const;
```
caiotheodoro, line 27: bug: this list also omits `action_intent` and `event_visibility`, both emitted on main (`pulse-scheduler.ts`). Add them back on rebase.
```suggestion
export const OPERATOR_EVENT_TYPES = [
  "llm_failure",
  "llm_budget_exceeded",
  "intent_blocked",
  "intent_delayed",
  "stagnation_warning",
  "scheduler_error",
  "pulse_metrics",
  "agent_state_snapshot",
  "action_intent",
  "event_visibility",
] as const;
```

caiotheodoro (changes requested):
Don't merge as-is: conflicting, and the union drops three types main already emits (`action_intent`, `event_visibility`, `agent_state_snapshot`). Two inline. Rebase, keep those three, drop only `rate_limit_hit`.

packages/shared/src/operator/operator.types.ts
Elesiann: Addressed in the current head (6a03bcc): `agent_state_snapshot` is kept in `OPERATOR_EVENT_TYPES` — this review was against the pre-rebase commit that dropped it. It is emitted from `persistAndSnapshot` in `pulse-scheduler.ts` (2 emission sites), and the producer guard now validates it alongside `action_intent`/`event_visibility`. Only `rate_limit_hit` is removed (the one declared type with no producer anywhere in `packages/server/src`).

packages/shared/src/operator/operator.types.ts
Elesiann: Addressed: the rebase against current main (which now carries the observability stream via #100) restored all three stream types — `action_intent`, `event_visibility` and `agent_state_snapshot` are in `OPERATOR_EVENT_TYPES` with emission sites in `pulse-scheduler.ts`, and the guard passes for each. The only member dropped is `rate_limit_hit`.

packages/server/src/simulation/__tests__/operator-event-producers.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  it.each(OPERATOR_EVENT_TYPES)(
+    "declared operator type %s has an emission site in packages/server/src",
+    (type) => {
+      expect(
+        allSource.includes(`type: "${type}"`),
+        `${type} is declared in shared but never emitted in packages/server/src — remove it from OPERATOR_EVENT_TYPES or wire a producer`,
+      ).toBe(true);
+    },
+  );
```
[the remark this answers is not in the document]
Elesiann: Valid concern — fixed in cc4f6fa. The guard no longer does a bare substring match: it now requires a `detail` property near the `type: "<name>"` literal. That is structural, not heuristic: `detail` is a required field of `OperatorEvent` and never appears on `SimulationEvent`, so committed-event constructions that share a type name (`intent_resolver` builds `type: "intent_blocked"`/`"intent_delayed"`, `pulse-scheduler` builds a committed `llm_failure`) cannot satisfy it. Both the explicit `detail:` and the shorthand `detail,` forms are matched (the `schedulerError` helper uses shorthand).

Added three regression tests, including the exact false-positive you described — a committed `SimulationEvent` shape with `type: "llm_failure"` (no `detail`) is asserted NOT to count, a real operator construction IS asserted to count, and the shorthand-detail form is pinned. Full gate green: 109 files / 1179 tests, audit 0 violations (node:22 container, `--frozen-lockfile`).

PR #102, opened 2026-08-26
feat(docker): containerized app runtime + tag-triggered GHCR release (proposal)

Elesiann:
## What

Three additive artifacts that let the simulation runtime run without a local Node 22 + pnpm 10 toolchain. No application code, no tests, and no new dependencies are touched.

1. **`docker/app/Dockerfile`** — multi-stage image. Stage 1 reproduces the CI order on `node:22-bookworm` (`pnpm install --frozen-lockfile` → `pnpm -r build`). Stage 2 installs only the production dependency graph of `@perfectman/server` (+ its workspace deps) and ships compiled `dist` trees. The entrypoint is the exact CLI that `pnpm --filter @perfectman/server simulation` runs natively — config discovery is untouched (walk-up finds `/app/config/index.json`, dotenv finds a mounted `/app/.env`).
2. **`docker/app/app.compose.yml`** — compose service following the existing `docker/` precedents (base file, combinable with `qwen3.compose.yml` via `-f -f`). Mounts `./config` read-only.
3. **`.github/workflows/docker-release.yml`** — on `v*` tag pushes, builds and publishes images to GHCR with `<major>.<minor>` tags from the tag name plus `latest` for non-prereleases.

Plus doc sections in both READMEs.

## What this flow enables

- **Zero-toolchain simulation**: clone → copy one example config into `config/` → `docker compose -f docker/app/app.compose.yml up --build`. Personas converse on stdout with zero API keys (mock provider). Works on machines with no Node/pnpm at all.
- **Deployable runtime**: any Docker host can run a released image (`ghcr.io/<owner>/perfectman:<tag>`), and each image is traceable to the exact commit it was built from via OCI labels — no dev checkout, no drift between "what I tested" and "what runs".
- **Foundation for frontend consumers**: the runtime already has pluggable transports (Socket.IO among them), so serving a UI becomes a *deployment decision* layered on this image rather than a new toolchain problem. This PR deliberately does not open or expose anything network-facing — it just makes the runtime something you can pull and run anywhere, which is the precondition for whatever consumes it next.
- **Images that don't rot by convention**: every `v*` tag rebuilds and republishes; reproducibility-conscious consumers pin `:<major>.<minor>` instead of `latest`.

## Proposal points (discussion wanted here)

This doubles as a proposal; reshape any of it in threads.

- **Tag-as-release convention**: this repo has never had releases or tags. The workflow proposes the minimal convention — pushing `v*` is what constitutes a release. Alternative would be GitHub Releases-driven or manual dispatch; happy either way, tag-push seemed smallest.
- **GHCR namespace**: images land under `ghcr.io/<repo-owner>/perfectman` via `${{ github.repository }}`, so they follow whoever merges first on upstream.
- **Base image**: `node:22-bookworm` full (matches CI's Node 22; avoids native-build surprises for `better-sqlite3`) vs `slim`/`alpine` for smaller images. Bookworm chosen for safety on the native dep; size optimization can come later if anyone cares.
- **Runtime graph**: stage 2 ships only `@perfectman/server`'s production deps (+ workspace deps) instead of the whole workspace install. Open to shipping more/less depending on how eval wants to relate to images later (left out of scope here).

## Validation

- Full gate green on a clean `node:22` container mirroring CI order (`install → build → lint → test:all`, including the test-hygiene audit): all pass.
- Image built from this Dockerfile and executed against `examples/simulations/mock.inline-personas.example.json` mounted at `/app/config`: live simulation stream (agent messages, mock LLM cognition calls ~50ms, engine state snapshots evolving, agents creating a private channel among themselves mid-run), clean SIGTERM shutdown (`simulation_stopped`). Zero API keys involved.
- Release workflow exercised end-to-end before proposing: test tag pushed on the author's fork → all steps green → `ghcr.io/elesiann/perfectman:0.0.0-docker-probe.1` published (and correctly *no* `latest` generated, since the probe tag carried a prerelease suffix). Probe tag deleted afterwards.
- Build-invariant note found while validating: composite tsc projects silently skip `dist` emission when a stale `tsconfig.tsbuildinfo` rides in the build context (Dockerignore patterns don't cross path separators like gitignore does). The Dockerfile clears `packages/*/dist` and `packages/*/*.tsbuildinfo` after COPY, so builds are context-independent.

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -1,0 +1,44 @@
+name: Docker release
+
+on:
+  push:
+    tags: ["v*"]
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish-image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - uses: docker/setup-buildx-action@v3
+
+      - uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/${{ github.repository }}
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=raw,value=latest,enable=${{ !contains(github.ref_name, '-') }}
+
+      - uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: docker/app/Dockerfile
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/README.md
+++ b/README.md
@@ -89,6 +89,30 @@ pnpm freellm:dev        # API on :3001, dashboard on :5173
 
 Add provider keys and create a unified key via the dashboard, then set `FREELLMAPI_KEY` in `.env`.
 
+**Simulation in Docker** (no local Node/pnpm toolchain needed):
+
+```bash
+mkdir -p config
+cp examples/simulations/mock.inline-personas.example.json config/index.json
+docker compose -f docker/app/app.compose.yml up --build
+```
+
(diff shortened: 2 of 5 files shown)


yoarajota (approved):

Elesiann:
Recreated this PR as #107 — the fork it branched from was deleted by accident. Same branch (`feat/docker-runtime` @ `fc5675d`), same content and body; continuing discussion there.

PR #107, opened 2026-08-26
feat(docker): containerized app runtime + tag-triggered GHCR release (proposal)

Elesiann:
> Note: this PR replaces #102, which was closed automatically after its head fork was deleted by accident — it contains exactly the same changes and proposal.

## What

Three additive artifacts that let the simulation runtime run without a local Node 22 + pnpm 10 toolchain. No application code, no tests, and no new dependencies are touched.

1. **`docker/app/Dockerfile`** — multi-stage image. Stage 1 reproduces the CI order on `node:22-bookworm` (`pnpm install --frozen-lockfile` → `pnpm -r build`). Stage 2 installs only the production dependency graph of `@perfectman/server` (+ its workspace deps) and ships compiled `dist` trees. The entrypoint is the exact CLI that `pnpm --filter @perfectman/server simulation` runs natively — config discovery is untouched (walk-up finds `/app/config/index.json`, dotenv finds a mounted `/app/.env`).
2. **`docker/app/app.compose.yml`** — compose service following the existing `docker/` precedents (base file, combinable with `qwen3.compose.yml` via `-f -f`). Mounts `./config` read-only.
3. **`.github/workflows/docker-release.yml`** — on `v*` tag pushes, builds and publishes images to GHCR with `<major>.<minor>` tags from the tag name plus `latest` for non-prereleases.

Plus doc sections in both READMEs.

## What this flow enables

- **Zero-toolchain simulation**: clone → copy one example config into `config/` → `docker compose -f docker/app/app.compose.yml up --build`. Personas converse on stdout with zero API keys (mock provider). Works on machines with no Node/pnpm at all.
- **Deployable runtime**: any Docker host can run a released image (`ghcr.io/<owner>/perfectman:<tag>`), and each image is traceable to the exact commit it was built from via OCI labels — no dev checkout, no drift between "what I tested" and "what runs".
- **Foundation for frontend consumers**: the runtime already has pluggable transports (Socket.IO among them), so serving a UI becomes a *deployment decision* layered on this image rather than a new toolchain problem. This PR deliberately does not open or expose anything network-facing — it just makes the runtime something you can pull and run anywhere, which is the precondition for whatever consumes it next.
- **Images that don't rot by convention**: every `v*` tag rebuilds and republishes; reproducibility-conscious consumers pin `:<major>.<minor>` instead of `latest`.

## Proposal points (discussion wanted here)

This doubles as a proposal; reshape any of it in threads.

- **Tag-as-release convention**: this repo has never had releases or tags. The workflow proposes the minimal convention — pushing `v*` is what constitutes a release. Alternative would be GitHub Releases-driven or manual dispatch; happy either way, tag-push seemed smallest.
- **GHCR namespace**: images land under `ghcr.io/<repo-owner>/perfectman` via `${{ github.repository }}`, so they follow whoever merges first on upstream.
- **Base image**: `node:22-bookworm` full (matches CI's Node 22; avoids native-build surprises for `better-sqlite3`) vs `slim`/`alpine` for smaller images. Bookworm chosen for safety on the native dep; size optimization can come later if anyone cares.
- **Runtime graph**: stage 2 ships only `@perfectman/server`'s production deps (+ workspace deps) instead of the whole workspace install. Open to shipping more/less depending on how eval wants to relate to images later (left out of scope here).

## Validation

- Full gate green on a clean `node:22` container mirroring CI order (`install → build → lint → test:all`, including the test-hygiene audit): all pass.
- Image built from this Dockerfile and executed against `examples/simulations/mock.inline-personas.example.json` mounted at `/app/config`: live simulation stream (agent messages, mock LLM cognition calls ~50ms, engine state snapshots evolving, agents creating a private channel among themselves mid-run), clean SIGTERM shutdown (`simulation_stopped`). Zero API keys involved.
- Release workflow exercised end-to-end before proposing: test tag pushed on the author's fork → all steps green → `ghcr.io/elesiann/perfectman:0.0.0-docker-probe.1` published (and correctly *no* `latest` generated, since the probe tag carried a prerelease suffix). Probe tag deleted afterwards.
- Build-invariant note found while validating: composite tsc projects silently skip `dist` emission when a stale `tsconfig.tsbuildinfo` rides in the build context (Dockerignore patterns don't cross path separators like gitignore does). The Dockerfile clears `packages/*/dist` and `packages/*/*.tsbuildinfo` after COPY, so builds are context-independent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,30 @@ pnpm freellm:dev        # API on :3001, dashboard on :5173
 
 Add provider keys and create a unified key via the dashboard, then set `FREELLMAPI_KEY` in `.env`.
 
+**Simulation in Docker** (no local Node/pnpm toolchain needed):
+
+```bash
+mkdir -p config
+cp examples/simulations/mock.inline-personas.example.json config/index.json
+docker compose -f docker/app/app.compose.yml up --build
+```
+
+It combines with the auxiliary containers, e.g. Qwen3/Ollama alongside the runtime:
+
+```bash
+docker compose -f docker/qwen3/qwen3.compose.yml \
+               -f docker/app/app.compose.yml up -d simulation
+```
+
+Published images are built for each tagged release and pushed to GHCR (`ghcr.io/caiotheodoro/perfectman:<tag>`) — see [Releases](#releases) below.
+
+This builds a multi-stage image of the runtime itself and runs the same CLI as `pnpm --filter @perfectman/server simulation`. With the mock example it needs no API keys or model — personas converse on stdout. To use a real provider, point your mounted `config/index.json` at it and hand keys to the container with an extra read-only mount of the same `.env` the CLI reads natively:
+
+```bash
+docker compose -f docker/app/app.compose.yml run --rm \
+  -v "$PWD/.env:/app/.env:ro" simulation
+```
+
 ### Bring your own personas
 
 See [`docs/personas/README.md`](docs/personas/README.md) for the plug-and-play persona interview/compile workflow. Real, person-specific persona files live under `config/` and `docs/personas/`, both gitignored — nothing personal gets committed.
@@ -105,6 +129,18 @@ The full design lives in [`docs/README.md`](docs/README.md) — start there for 
 ## Status / open questions
 
 This is an active experiment, not a finished product. See [`docs/README.md`](docs/README.md#current-open-questions) for what's genuinely undecided — private-channel spectator visibility, how much scoring machinery should exist, which symbolic actions are worth keeping. If one of those interests you, that's a good place to start.
+
+## Releases
+
+> Proposal — this convention only exists once this lands and the first tag is cut.
+
+Every tagged release (`v*`) publishes a container image of the runtime to GitHub Container Registry via `docker-release.yml`: `ghcr.io/caiotheodoro/perfectman` gets the exact tag (`v0.1.0` → image tag `0.1.0`, plus `latest` for non-prerelease). Consumers without a Node toolchain can run the simulation directly:
+
+```bash
+docker pull ghcr.io/caiotheodoro/perfectman:latest
+```
+
+Use a tagged image instead of `latest` anywhere reproducibility matters.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -97,11 +97,13 @@ cp examples/simulations/mock.inline-personas.example.json config/index.json
 docker compose -f docker/app/app.compose.yml up --build
 ```
 
-It combines with the auxiliary containers, e.g. Qwen3/Ollama alongside the runtime:
+It combines with the auxiliary containers — putting the app file first keeps every relative path anchored to `docker/app/`. With Ollama running as an in-project service, point your config's provider `baseURL` at `http://qwen3:11434/v1` (container DNS), not localhost:
 
 ```bash
-docker compose -f docker/qwen3/qwen3.compose.yml \
-               -f docker/app/app.compose.yml up -d simulation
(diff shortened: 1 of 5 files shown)


docker/app/app.compose.yml
```diff
@@ -0,0 +1,19 @@
+name: perfectman
+
+services:
+  simulation:
+    build:
+      context: ..
+      dockerfile: app/Dockerfile
```
caiotheodoro: bug: compose paths are relative to `docker/app/`, so `context: ..` is `docker/` and `COPY package.json` fails. Match `freellmapi.compose.yml`: `context: ../..`, `dockerfile: docker/app/Dockerfile`.
```suggestion
      context: ../..
      dockerfile: docker/app/Dockerfile
```

README.md
```diff
@@ -89,6 +89,30 @@ pnpm freellm:dev        # API on :3001, dashboard on :5173
...
+mkdir -p config
+cp examples/simulations/mock.inline-personas.example.json config/index.json
+docker compose -f docker/app/app.compose.yml up --build
+```
+
+It combines with the auxiliary containers, e.g. Qwen3/Ollama alongside the runtime:
+
+```bash
+docker compose -f docker/qwen3/qwen3.compose.yml \
+               -f docker/app/app.compose.yml up -d simulation
```
caiotheodoro: bug: `up -d simulation` never starts the `qwen3` service, so this is not alongside. Drop the service filter (`up -d`). In-cluster Ollama is `http://qwen3:11434`, not `host.docker.internal`.

caiotheodoro (changes requested):
Don't merge until the compose build context is fixed — the documented `up --build` copies `docker/` as the repo root. Two inline.

docker/app/app.compose.yml
Elesiann: Real bug — thanks. Applied your suggestion verbatim (`context: ../..` + explicit `dockerfile`). Verified against the documented flow itself: `docker compose -f docker/app/app.compose.yml up --build` now builds and runs the mock simulation through compose (259 lines of stream, `agent_message` events, clean container stop); the combined-stack `config` resolves `context:` to the repo root and `dockerfile: docker/app/Dockerfile`. (My earlier smoke tests always built the image directly instead of going through compose — which is exactly how this slipped through.)

README.md
Elesiann: Agreed on all three points. Rewrote that snippet: dropped the service filter, put the app file first in the `-f` chain so the relative anchors stay stable regardless of what comes after it, and documented the in-project Ollama URL as `http://qwen3:11434/v1` (container DNS). Kept the `host.docker.internal:host-gateway` extra_host on the service for the native-Ollama-on-host case, with the config note updated accordingly.

Elesiann:
Both threads addressed in `5753640`: build context fixed exactly as suggested (`context: ../..` + explicit `dockerfile`), and the combined-stack docs now run app-file-first with both services unfiltered plus the in-project Ollama baseURL note. Validated through the documented flows themselves this time — `up --build` drives the mock simulation end-to-end via compose, and the merged `config` resolves `context:`/config-mount paths from the repo root. Ready for re-review.

README.md
caiotheodoro: right — unfiltered `up -d` with the app file first, and `http://qwen3:11434/v1` for in-cluster Ollama.

docker/app/app.compose.yml
caiotheodoro: right, `context: ../..` + `dockerfile: docker/app/Dockerfile` is the same shape as freellmapi.compose.yml.

Issue #106, opened 2026-08-26
Calibrate the meaning-made 0.33 gate (world-evaluator.ts:643)


yoarajota:
## Summary

From the #96 goal-layer calibration (mock-only sweep, 2026-08-26), the only measured near-threshold in the goal layer is the hardcoded meaning-made gate `divergenceFromLog < 0.33` in `deriveMeaningMade` (packages/server/src/simulation/world/world-evaluator.ts:643). It sits outside the #96 sweep matrix (RD-4: weights x reviewEveryPulses x offerAcceptPulses), so it was intentionally NOT tuned there — tuning without evidence was rejected (ADR-0011 D-28). This issue tracks its dedicated calibration.

## Measured evidence

From docs/research/goal-layer/calibration-2026-08-26.md (sec 7) and evidence/goal-trajectory-sweep-2026-08-26.json:

- Closest approach to the gate on any measured sample: the healthy arc's pre-flip continue review — divergenceFromLog 0.326 at reviewEveryPulses=1 (1.2% margin under the 0.33 gate), 0.316 at cadence 10 (4.4%).
- Flip reviews that actually consulted the gate: healthy arc 0.253 at reviewEveryPulses=1 (23% margin), healthy arc 0.300 at cadence 10 (9.1%), world-briefly-wrong 0.307 (7.0% — the tightest gate-decisive margin).
- The contested window on the world-briefly-wrong arc measures 0.376 and correctly fails the gate (no false accept).
- The gate is server-side, hardcoded, and has no config path today.

Margin definition: (0.33 - divergenceFromLog) / 0.33. The tightest margin on any measured sample was 1.2% (a pre-flip continue review on the healthy arc at cadence 1); the tightest margin among flip reviews that consulted the gate was 7.0% (world-briefly-wrong).

## Follow-up sweep outline

- Elevate the constant to the `goalLayer` config section (optional field, default 0.33) or sweep a small grid (e.g. 0.25 / 0.30 / 0.33 / 0.36 / 0.40) over the healthy-achiever, world-briefly-wrong, and hollow-completion arcs at cadence 1 and 10.
- Success signal: false-reject rate (healthy/world-briefly-wrong arcs failing the gate) vs false-accept rate (contested windows passing it), plus a target flip-margin floor.
- Reuse the #96 harness (mock-only, zero wire calls, deterministic trajectories; pulse cap 120).

Related: #96, ADR-0011 (docs/adr/0011-goal-layer-threshold-calibration.md).

Issue #98, opened 2026-08-25
Docs research-readiness: lint step, source backfill, concept pages (R4 follow-through)


yoarajota:
## Context

The docs audit vs Karpathy's llm-wiki (#93 research R4) found `docs/` only partially
research-ready. This run fixed the pattern for NEW artifacts (docs/research/ packs, inline ADR
rationale, research log). This issue tracks the backlog repair the audit named:

## Scope

1. **Backfill dead citations**: `docs/architecture/emotion.md` cites Russell's Circumplex with no
   URL; `docs/notes/design-conversation-history.md` has bare `ScienceDirect` / `arxivUpenn` cites
   (no URL/title/date). Move to a `docs/research/` pack with a source-map or add inline `## Sources`.
2. **Docs lint step**: wired into the pipeline — dead relative links, archived-path references in
   committed docs, orphan concept pages, uncited claims, stale dates (precedent:
   `docs/eval/README.md` "Calibration-pending").
3. **Concept pages**: stop appending to `docs/concepts/concept-map.md` (29 concepts, 1372 lines);
   new concepts are atomic pages + catalog row (the goal-layer pattern). Optionally split historic
   concepts later.
4. **ADR template tweak**: new ADRs embed Rationale + Rejected Alternatives inline (done for
   0008/0009; document the convention).

## Acceptance criteria

- [ ] emotion.md + design-conversation-history.md citations re-derivable (URL/type/date/verification)
- [ ] Docs lint runs in the pipeline gate, first pass clean (or explicit waivers recorded)
- [ ] Convention documented (research pack layout + ## Sources footer + ADR inline rationale)

References: #93 (R4), `docs/research/goal-layer/` (the pattern this issue generalizes),
`docs/notes/discord/source-map.md` (in-repo template).

---
## Run context (fresh session)

- **Baseline branch**: `feat/emergent-goal-layer` @ `8d298df` — `docs/research/goal-layer/` (the pattern to generalize) exists ONLY there; PR pending. The files to repair (`docs/architecture/emotion.md`, `docs/notes/design-conversation-history.md`) are on `main`.
- **Decision IDs** resolve in the committed `docs/adr/0009-goal-layer-runtime-wiring.md` (D-3) — pipeline records are not committed.

yoarajota:
landed via PR #110 (merge 0ccdff7) — dead citations re-derived, docs-lint wired into the gate, conventions documented incl. the LLM-Wiki lineage.

Issue #97, opened 2026-08-25
Persist goal-layer state across restarts (sqlite registry backing)


yoarajota:
## Context

#93 shipped the `GoalRegistry` in-memory only (ADR-0009 D-4: in-memory Map + pure replay reducer
over committed world events). Restart continuity is currently covered by replay at near-zero cost;
durability is not a requirement of the V1 slice. This issue tracks the optional upgrade.

## Scope

1. Evaluate what must persist: registry state (proposals, accepted goals, verdicts, gap history,
   pending offers) vs what replay already reconstructs from committed events.
2. If a gap exists: sqlite backing behind the registry following the repo's persistence pattern
   (`packages/server/src/persistence/`), keeping the replay reducer as the source of truth and the
   DB as an accelerator/cache.

## Acceptance criteria

- [ ] Replay-from-events and sqlite-backed state agree on the same committed log (parity test)
- [ ] Restart of a mid-run simulation resumes goals/verdicts/gap history

Priority: low (optional). References: #93, ADR-0009 (D-4), `packages/server/src/persistence/`.

---
## Run context (fresh session)

- **Baseline branch**: `feat/emergent-goal-layer` @ `8d298df` — the world/ subsystem and goal contracts exist ONLY there; PR pending. Run from that branch (or after merge to main).
- **Decision IDs** resolve in the committed `docs/adr/0009-goal-layer-runtime-wiring.md` (D-4) — pipeline records are not committed.
- **Invariant**: the replay reducer stays the source of truth; any sqlite backing must pass parity with it.

yoarajota:
landed via PR #109 (merge f8887a5) — sqlite-backed junction persistence with replay parity; restart resumes selfVerdicts.

Issue #96, opened 2026-08-25
Calibrate goal-layer thresholds via mock-run experiments


yoarajota:
## Context

The goal layer's V1 thresholds are scaffold defaults, flagged in
`docs/research/goal-layer/gaps.md`: `DEFAULT_DELUSION_WEIGHTS` (wSignal 0.4 / wSocial 0.4 /
wIdentity 0.2 / revisionThreshold 0.5), crystallizer trigger minima (2 blocked intents, 3 gaze
events), `reviewEveryPulses` default 10, `offerAcceptPulses` default 0. No empirical basis yet.

## Scope

1. Scenario matrix (mock provider, free): healthy achiever, deluded achiever, premature closer
   (public acknowledgment), contested consensus (in-group vs collective), hollow completion,
   world-briefly-wrong; 4-persona setups from the e2e fixtures.
2. Capture per run: goals crystallized/accepted, gap magnitude trajectories, termination reasons
   (reached / story-is-over / re-goal), pulse counts.
3. Tune thresholds against desired behavior; record the calibration run in `docs/research/goal-layer/`
   (freshness-dated, per the source-map convention).

## Acceptance criteria

- [ ] Trajectory data for ≥6 scenarios committed under `docs/research/goal-layer/`
- [ ] Defaults updated from evidence (or explicitly kept with the negative result documented)
- [ ] No LLM cost (mock provider only)

References: #93, `packages/engine/src/goal/` (threshold constants),
`docs/research/goal-layer/gaps.md`, `docs/eval/` (calibration-pending convention).

---
## Run context (fresh session)

- **Baseline branch**: `feat/emergent-goal-layer` @ `8d298df` — the goal-layer feature (packages/server/src/simulation/world/, packages/shared/src/goal/, ADR-0008/0009, docs/research/goal-layer/) exists ONLY there; its PR is still pending. Run from that branch (or after merge to main).
- **Decision IDs** (D-###) above resolve in the committed `docs/adr/0009-goal-layer-runtime-wiring.md`. Pipeline records under `.claude/_output/` are NOT committed — the ADR is the durable home.
[...]

PR #109, opened 2026-08-27
feat(server): sqlite-backed goal-registry persistence with replay parity (#97)

yoarajota:
## What

Optional sqlite backing for the goal registry (#97): the replay reducer over committed world events stays the source of truth, and the DB holds only what replay cannot rebuild — the `selfVerdicts` junction. Persistence derives from the existing `persistence.type` (no new config knob).

## Changes

- **Schema + repo**: `goal_self_verdicts` table (keyed simulation_id, goal_id) + `SqliteGoalRegistryRepository` (memory-repository shape; **no `IGoalRegistryRepository`** — ADR-0009 D-1 stays), exported from the server package
- **Write path (D-31)**: whole-junction write-through at review end, after `appendAndProject`, every review including candidate-less LLM intervals; persist failures contained with a distinct operator event — never fails a pulse
- **Restart path (T412)**: `attachExisting` get-or-create on `SimulationRuntime.createSimulation` (new option on `BuildConfiguredSimulationOptions`) — a mid-run restart through the supported factory no longer crashes on the simulations PK (this was the validator-blocked gap: AC-2 was unprovable without it); memory and sqlite modes converge on restart behavior
- **D-23 flip (D-35)**: junction restore on registry construction — sqlite-mode restarts resume selfVerdicts (documented degradation removed in sqlite mode only); termination still gates on the world verdict only (G5 intact)
- **Tests (AC-1/AC-2)**: parity test (one committed log → replay-built vs sqlite-backed state, value-level junction comparison) + file-backed restart test through the factory with the deluded-LLM leg as the strict mid-run proof (D-36)
- **ADR-0012** (D-29..D-36 LOCKED) + prose-only amendments to ADR-0009 (D-4 memory-mode status, D-1 superseded note) and ADR-0010 (D-23 consequence scoped to memory mode); concepts marker; inventory regenerated

## Acceptance criteria

- [x] Replay-from-events and sqlite-backed state agree on the same committed log (parity test)
- [x] Restart of a mid-run simulation resumes goals/verdicts/gap history (file-backed restart test, factory path, both persistence modes)

## Verification

- `pnpm test` 1244/1244 (wave-1 oracle wave: 1235 pass + 2 intended reds → green by wave 2); typecheck 4/4; audit gate PASS (113 files / 1096 its, inventory regenerated)
- Red-oracle discipline: T401/T402 written RED in wave 1 (module-load + PK-constraint failures), flipped GREEN only by the real implementation
- G4 (`packages/engine/src/goal/*`) and D-18 (`packages/server/src/llm/*`) diff-empty; zero new dependencies (better-sqlite3 already a server dep)
- Review: APPROVE (0 findings; 8 dropped)

## Notes

- Decision-ID queue state resolved: this run's ADR-0012 stamps D-29..D-36; the unmerged #95 (D-24..D-26, ADR-0009) and #96 (renumbered D-27..D-28, ADR-0011) own the preceding ranges — reconciliation precondition verified at write time
- Scope answer recorded in ADR-0012: the five committed shapes (proposals/goals/verdicts/gap history/pending offers) are fully reconstructed by replay — only `selfVerdicts` needed persistence
- Follow-ups still open: #98 (docs research-readiness), #106 (0.33 gate calibration)


packages/server/src/simulation/world/world-evaluator.ts
```diff
@@ -30,6 +30,7 @@ import {
 } from "@perfectman/engine";
 import type { DeferenceSignal, GoalRatingContext, WorldStateSnapshot } from "@perfectman/engine";
 import type { IAgentStateRepository, IEventRepository } from "../../persistence/repositories.js";
+import type { GoalSelfVerdictEntry } from "../../persistence/sqlite/goal-registry-repository.js";
```
caiotheodoro: risk: `GoalSelfVerdictEntry` is imported from the sqlite impl, and the D-32 comment on `GoalRegistryPersister` says world never does that. Define the entry type next to the persister and have the sqlite repo import it.

packages/server/src/persistence/sqlite/goal-registry-repository.ts
```diff
@@ -0,0 +1,73 @@
...
+        .run(
+          simulationId,
+          entry.goalId,
+          JSON.stringify(entry.verdict),
+          entry.source,
+          Date.now(),
+        );
+    }
+    return Promise.resolve();
+  }
```
caiotheodoro, line 58: bug: this never deletes. `saveSelfVerdicts(id, [])` is a no-op, so a dropped self-verdict is resurrected on restart. Wrap `DELETE FROM goal_self_verdicts WHERE simulation_id = ?` plus the inserts in one `db.transaction`.
```suggestion
  saveSelfVerdicts(
    simulationId: string,
    entries: GoalSelfVerdictEntry[],
  ): Promise<void> {
    const persist = this.db.transaction(() => {
      this.db
        .prepare(`DELETE FROM goal_self_verdicts WHERE simulation_id = ?`)
        .run(simulationId);
      const insert = this.db.prepare(
        `INSERT INTO goal_self_verdicts
           (simulation_id, goal_id, verdict, source, updated_at)
         VALUES (?, ?, ?, ?, ?)`,
      );
      const now = Date.now();
      for (const entry of entries) {
        insert.run(
          simulationId,
          entry.goalId,
          JSON.stringify(entry.verdict),
          entry.source,
          now,
        );
      }
    });
    persist();
    return Promise.resolve();
  }
```

caiotheodoro (changes requested):
Two inline. Also conflicting with main (#105 already landed) — rebase before anything else.

packages/server/src/simulation/world/world-evaluator.ts
yoarajota: moved it. `GoalSelfVerdictEntry` now lives next to `GoalRegistryPersister` in world-evaluator.ts and the sqlite repo imports the type from there — world stays clean of persistence impls

packages/server/src/persistence/sqlite/goal-registry-repository.ts
yoarajota: nice catch — empty write-through was a no-op and dropped verdicts came back on restart. now deletes the whole junction then re-inserts inside one `db.transaction`, plus a regression test pinning the dropped-goal and `[]` cases

yoarajota:
rebased onto main (#105 landed) and both fixes are in — inline replies above. fresh-checkout verification: 1275/1275 tests, typecheck 4/4, audit gate clean.

Issue #129, opened 2026-08-30
Audit: conversation quality / emergent behaviour after relief + memory land


yoarajota:
Child of the [simulation runtime & behavior fix plan](https://github.com/caiotheodoro/perfectman/issues/119) map. **Type: grilling** (HITL). Graduated from #120 (degeneracy = "contributing, not root cause"). **Blocked by #124 (lastActionAt + relief) and #126 (memory).**

## Question

Once accumulator relief (#124) and a working memory subsystem (#126) are in place, does conversation quality recover — and if not, what drives the degeneracy?

The `lastActionAt` gap raised turn *rate*, not turn *quality*. After #124 / #126 land, re-run and audit what remains.

## Decisions to make

1. **Re-run first** — a wider validation run (varied archetypes, 3–4 agents, longer horizon) with #124 + #126 merged. Does emergent behaviour (exclusion, masking, grudges) appear?
2. **Prompt / persona salience** — is the cognition prompt giving the model enough differentiated signal (persona voice, relationship state, recent memory) to diverge, or is it structurally pushed toward bland consensus?
3. **Model choice** — is `deepseek-v4-flash` (temp 0.7, `thinking: disabled`) adequate, or does the run need a stronger model to show the target behaviours?
4. **Interaction with relational state (#127)** — how much of the flatness was the absence of that input.

## Resolves

Whether degeneracy is fixed by #124 / #126 or needs its own prompt / persona / model changes, backed by a fresh run.

yoarajota:
Stale

Issue #125, opened 2026-08-30
Decide: stagnation-detection rework (heuristic re-tune + wire detectAttractorStates + per-cadence telemetry)


yoarajota:
Child of the [simulation runtime & behavior fix plan](https://github.com/caiotheodoro/perfectman/issues/119) map. **Type: grilling** (HITL). Graduated from #120 (mostly-independent gap). Consult `docs/research/lastactionat-root-cause.md`.

## Question

How should stagnation detection be reworked so it catches a textbook conversational loop?

In the run, `computeStagnationMetrics` (`packages/engine/src/health/compute-stagnation-metrics.ts:29-221`) ran ~4x and never left `normal` while both agents looped on vague mutual agreement.

## Decisions to make

1. **Heuristic re-tune** — the metric reads events + relational trust + coreMood + accumulator-over-threshold counts, no `lastActionAt`. Blindness is RDV/ERI variance collapse plus an unbounded history window; universal accumulator saturation additionally drives IGE to ~0.06, masking stagnation. Decide the sensitivity changes, the history-window bound, and how to make IGE robust to the saturation being fixed in #119-T5.
2. **Wire `detectAttractorStates`** — it has the `message_loop` signature and is exported but never called by `pulse-scheduler.ts`. Wire it in, or fold its logic into `computeStagnationMetrics`.
3. **Telemetry** — emit the stagnation composite every cadence (`pulse-scheduler.ts:329` currently emits only when it trips), so a run's stagnation trend is observable.

## Resolves

The reworked heuristic, the attractor-detection wiring decision, and the telemetry cadence.

yoarajota:
## Resolution

Decided via grilling. The guiding call: **#125 makes only structural changes; all numeric re-tuning is deferred** until #124 (saturation) + #126 (memory) land and there's a clean run to tune against — tuning thresholds now would compensate for bugs we're about to fix.

### Decisions

| | |
|---|---|
| **Scope** | Structural only in #125: bound the history window, wire attractor detection, add per-cadence telemetry. Numeric re-tuning (RDV/ERI sensitivity, IGE formula robustness to the saturation fix, sub-metric weights, `yellow`/`red`/`critical` thresholds, window size N) → fog item, unblocked by #124 + #126 + a clean validation run. |
[...]

PR #139, opened 2026-08-30
feat(simulation): bound stagnation window, wire attractor detection (#133)

### LINKED ISSUE #133: Stagnation detection — structural: rolling window, wire detectAttractorStates, per-cadence telemetry
## Parent

Decision: #125 (stagnation-detection rework — structural changes only). Plan: #119.

## What to build

The stagnation detector notices a two-agent conversational loop, and its score trend is observable every cadence instead of only when it trips.

- Feed `computeStagnationMetrics` a rolling window of roughly the last 40 pulses' events instead of the full committed history.
- Wire `detectAttractorStates` into the pulse scheduler alongside `computeStagnationMetrics`; on a detected signature (`message_loop`, `single_topic_lock`) emit a distinct `attractor_detected` operator event naming the signature. It does not override the composite `level`.
- Emit the stagnation composite score and all seven sub-metrics as an operator event every stagnation cadence, regardless of `level`.
- No changes to weights, thresholds, or the RDV / ERI / IGE formulas — those are deferred (fog, blocked on #124 + #126 + a clean run).

## Acceptance criteria

- [ ] A synthetic run where two agents alternate near-identical messages raises an `attractor_detected` operator event.
- [ ] The metrics input is bounded to the rolling window (older events excluded), verified by test.
- [ ] Every stagnation cadence emits a composite + sub-metrics operator event even when `level` is `normal`.
- [ ] `level` is still computed exactly as before an attractor hit — no override.

## Blocked by

None (can start immediately).

yoarajota:
## What

Restructures stagnation detection in the pulse scheduler — no threshold, weight, or formula changes:

- Feeds `computeStagnationMetrics` a rolling window of committed events, `STAGNATION_WINDOW_PULSES = 40` (shared constant, raised from an unused `30`; the pulse scheduler no longer keeps a local copy), instead of the full `getCommittedThrough` history.
- Calls `detectAttractorStates` (exported from `compute-stagnation-metrics.ts` but never invoked) alongside the composite score and emits one `attractor_detected` operator event per detected signature (`message_loop`, `silence_cascade`, ...). It stays independent of the composite `level` and never overrides it.
- Emits `stagnation_metrics` — composite plus all seven sub-metrics — as an operator event every stagnation cadence regardless of `level`. `stagnation_warning` still fires only on a non-normal `level`.
- Adds `stagnation_metrics` and `attractor_detected` to `OPERATOR_EVENT_TYPES`; `AttractorDetectedOperatorData.signature` is the `AttractorState` union, validated against `ATTRACTOR_THRESHOLDS` at the emission site (`detectAttractorStates` stays `string[]` engine-side).
- Known, deferred to #129: an empty/sparse window (room silent for 40+ pulses) leaves the metrics on neutral no-data defaults that sum to `0.604`, already `>= STAGNATION_THRESHOLDS.yellow (0.60)`, so a cadence commits `stagnation_detected` off defaults not a real measurement. Thresholds are out of scope here (decision #125). A threshold-free fix would gate the composite commit on a minimum window-activity floor; not implemented unilaterally.
- 7 tests: per-cadence emit at `level: normal`, `attractor_detected` on a seeded alternating message-loop run, one event per detected signature, no `level` override on an attractor hit, metrics input bounded to the window (pulses `5/8/10` excluded at pulse 50), empty-window `0.604`/`yellow` pin, sparse-window non-`normal` pin.

## Why

`computeStagnationMetrics` was fed the entire committed history, so a lively opening stretch kept the composite score healthy long after a room went inert. `detectAttractorStates` shipped with `message_loop` / `silence_cascade` / `dormant_agents` / etc. signatures but had no call site, and the cadence only surfaced telemetry once `level` was already non-normal — leaving nothing to tune the detectors against.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+7)
- 119 files, 1324 tests passed; `pnpm build` + per-package `pnpm --filter … typecheck` (shared/engine/server) clean; `pnpm test:gate` PASS (audit-tests 1155 its, 0 violations)
- window bound proven end to end: at pulse 50 the detector input is `[12, 30, 49]`, pulses `5/8/10` excluded

Closes #133

### DIFF
--- a/packages/engine/src/__tests__/stagnation.test.ts
+++ b/packages/engine/src/__tests__/stagnation.test.ts
@@ -5,6 +5,7 @@ import type {
   CoreMood,
   SocialEmotions,
 } from "@perfectman/shared";
+import { STAGNATION_THRESHOLDS } from "@perfectman/shared";
 import {
   computeStagnationMetrics,
   detectAttractorStates,
@@ -144,6 +145,26 @@ describe("computeStagnationMetrics", () => {
     for (const key of ["bdi", "rdv", "ige", "cue", "eri", "isd", "cns"] as const) {
       expect(typeof result[key]).toBe("number");
     }
+  });
+
+  // KNOWN, deferred to #129. An empty rolling window — a room silent for the
+  // whole window — leaves every metric on its neutral no-data default, whose
+  // weighted sum is 0.604, already >= STAGNATION_THRESHOLDS.yellow (0.60). The
+  // pulse scheduler then commits stagnation_detected off defaults rather than a
+  // real measurement. Threshold tuning is out of scope for #133 (decision
+  // #125); this test pins the behavior so any fix is a deliberate change.
+  it("scores yellow from neutral defaults on an empty window", () => {
+    const result = computeStagnationMetrics("sim1", 50, [], new Map());
+    expect(result.compositeScore).toBeCloseTo(0.604, 5);
+    expect(result.compositeScore).toBeGreaterThanOrEqual(STAGNATION_THRESHOLDS.yellow);
+    expect(result.level).toBe("yellow");
+  });
+
+  it("scores at or above yellow on a sparse window with no agent states", () => {
+    const events = Array.from({ length: 2 }, (_, i) => makeEvent(`${i}`, "a1", "message_sent"));
+    const result = computeStagnationMetrics("sim1", 50, events, new Map());
+    expect(result.compositeScore).toBeGreaterThanOrEqual(STAGNATION_THRESHOLDS.yellow);
+    expect(result.level).not.toBe("normal");
   });
 });
 
--- a/packages/engine/src/health/compute-stagnation-metrics.ts
+++ b/packages/engine/src/health/compute-stagnation-metrics.ts
@@ -2,6 +2,7 @@ import type {
   CommittedEvent,
   AgentState,
   StagnationMetrics,
+  AttractorState,
 } from "@perfectman/shared";
 import {
   STAGNATION_THRESHOLDS,
@@ -222,13 +223,13 @@ function computeCNS(events: CommittedEvent[]): number {
 
 /**
  * Detect attractor states from recent events.
- * Returns array of detected attractor state labels.
+ * Returns the detected signatures; an empty array means none fired.
  */
 export function detectAttractorStates(
   recentEvents: CommittedEvent[],
   agentStates: Map<string, AgentState>,
-): string[] {
-  const detected: string[] = [];
+): AttractorState[] {
+  const detected: AttractorState[] = [];
   const WINDOW = ATTRACTOR_DETECTION_WINDOW_PULSES;
 
   const windowEvents = recentEvents.slice(-WINDOW * 3); // rough pulse window
--- a/packages/server/src/simulation/__tests__/fixtures.ts
+++ b/packages/server/src/simulation/__tests__/fixtures.ts
@@ -1,0 +1,109 @@
+/**
+ * Shared fixtures for PulseScheduler tests.
+ * Defaults target the stagnation-window suite ("agent_1" in "sim_stag");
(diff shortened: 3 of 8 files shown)


matheusht:
**Two-axis review** (per the [code-review skill](https://github.com/mattpocock/skills/blob/main/skills/engineering/code-review/SKILL.md)): two independent agents reviewed `main...ad40de8` — **Standards** (repo standards + Fowler smell baseline) and **Spec** (issue #133). Static review; axes kept separate, no reranking. No tests were run by the reviewers.

## Standards
0 hard violations; 4 judgement calls:

1. **Q5 fixture duplication** — `pulse-scheduler-stagnation-window.test.ts:46-160` clones `SETTINGS`, `makeAgentState`, `makePersona`, `cannedNoOpStep`, `ZERO_ACTION` from `pulse-scheduler.test.ts`. Nine sibling tests do the same (no shared simulation fixture module exists), so this extends existing convention; the real fix is a `simulation/__tests__/fixtures.ts`, out of scope here.
2. **Duplicated Code** — `metricsData` (`pulse-scheduler.ts:338-350`) hand-copies all ten fields of `StagnationMetrics` although `StagnationMetricsOperatorData` is an alias of it (`data: metrics` would be type-identical). Ten drift points per new metric. Counter-argument: an explicit copy pins the operator payload shape against engine-side additions.
3. **Primitive Obsession-adjacent** — `string[] → AttractorState` narrowed at the consumer (`pulse-scheduler.ts:361-365`: filter + cast) instead of typing `detectAttractorStates`' return at the source. Documented inline.
4. **Duplicated Code** — the metric-key tuple `["bdi", "rdv", ...]` appears verbatim in both new tests (~559 observability, ~265 stagnation-window).

## Spec
All four acceptance criteria implemented (window bound test-proven: detector input `[12, 30, 49]` at pulse 50). Two findings:

1. **(P2) Bounded window newly makes a fully silent room score yellow off no-data defaults.** Empty window → all seven metrics fall to neutral defaults summing to `0.604` ≥ `STAGNATION_THRESHOLDS.yellow` (0.60) → the cadence commits `stagnation_detected` from defaults, not measurement. Declared known in the PR body and deferred to #129 per decision #125, with pinning tests — consistent with the spec's no-threshold-change clause, but the false-positive surface is real and now reachable after any lively start goes quiet.
2. **(P3) `single_topic_lock` can never be emitted.** Spec: "on a detected signature (`message_loop`, `single_topic_lock`) emit a distinct `attractor_detected`". `detectAttractorStates` (`compute-stagnation-metrics.ts:227`) implements no `single_topic_lock` branch — pre-existing engine gap (file untouched in this diff).

---
**Per axis:** Standards — 4 findings (worst: `metricsData` field-clone, P3) · Spec — 2 findings (worst: silent-room defaults → false `stagnation_detected`, P2, deferred to #129).

yoarajota:
@matheusht consolidated the fixtures in `simulation/__tests__/fixtures.ts` and exported STAGNATION_METRIC_KEYS once, detectAttractorStates returns AttractorState[] now so the consumer cast is gone. the silent-room yellow stays deferred to #129 like agreed in #125, and single_topic_lock is a pre-existing engine gap, left it out of scope. metricsData copy stays — it pins the operator payload shape against engine-side additions

caiotheodoro:
tl;dr: Good instrumentation. One observation on the known-defaults behavior.

- Bounding the stagnation window with the rolling-event approach prevents unbounded memory growth in long simulations.
-  returning  instead of  is a good type narrowing — the union is closed and self-documenting.
- The pulse-scheduler test (9 seed messages, 11 pulses) validates that  emits per-cadence and  emits as a distinct event — the two don't interfere.
- The known-defaults test (scores yellow from empty window) pins the current behavior so the threshold tuning in #129 is a deliberate change. Good practice.

caiotheodoro (approved):
LGTM

PR #141, opened 2026-08-30
fix(server): resolve reply/reaction targets via event handles (#135)

### LINKED ISSUE #135: Reply/reaction target resolution: ordinal event handles + parser resolve + retry + triggering-event floor
## Parent

Decision: #123 (reply/reaction target resolution). Plan: #119.

## What to build

A reply or reaction reliably points at the real event it responds to, and the model is given a way to name that event.

- Each event shown to the model in the prompt is prefixed with a short per-render handle (`[e1]`, `[e2]`, …); the perception packet carries a `handle → real event id` map for that render.
- The intent parser resolves the model's `replyToEventId` / `targetEventId` handle to the real event id.
- If a `reply_to_message` or `react` intent has an unresolvable target, that is a retriable validation failure: the existing action-intent retry arm re-prompts once (budget permitting) with a correction note that names the bad handle and lists the valid ones.
- If the retry is still unresolvable: infer the triggering event as the target (for both replies and reactions). Only when there is no triggering event and no visible message: downgrade the reply to a plain message, drop the reaction. Never commit an empty or unresolved target.
- The resolved event's actor is written as `replyToActorId` on the `reply_sent` payload.

## Acceptance criteria

- [ ] A reply that references a valid handle commits with `replyToEventId` set to the real event id and `replyToActorId` set to that event's actor.
- [ ] A reply with a bogus handle triggers exactly one retry with a correction note, then falls to the triggering-event floor.
- [ ] A reaction with no resolvable target and no triggering event is dropped, not committed with an empty `targetEventId`.
- [ ] No committed `reply_sent` / `reaction_sent` has an empty or non-existent target id.

## Blocked by

None (can start immediately).

yoarajota:
## What

Gives the model a stable way to point at a specific event and resolves it engine-side:

- `buildPerceptionPacket` emits a per-render `eventHandles` map (`e1` -> real `evt_*` id) covering the triggering event and every visible-context event; one handle per distinct id.
- `ActionIntentPromptBuilder.formatEvent` prefixes each shown event with its `[e1]` handle, and `<output_contract>` / `<events>` instruct the model to set `replyToEventId` / `targetEventId` to a handle, verbatim.
- `IntentParser.parse` takes a `TargetResolutionContext`, resolves the handle (or a bracket-wrapped / verbatim real id) back to the real event id after schema validation, and stamps that event's actor as `replyToActorId`. An unresolvable handle returns `unresolvedTarget` (retriable), not a `no_op` fallback.
- `action-intent-step` re-prompts once with a targeted note naming the bad handle, the valid handle set, and the `send_message` escape hatch; then `IntentParser.floorTargets` infers the triggering event (else the most recent visible message) as the target for both `reply_to_message` and `react`. With neither, the reply downgrades to `send_message` and the reaction is dropped. `replyToEventId` / `targetEventId` is never committed empty or unresolved.
- `buildReplyEvent` writes `replyToActorId` into the `reply_sent` payload; `deriveFallbackIntent` carries it into a reply fallback.
- 23 tests: handle/bracket/verbatim-id resolution, invented-slug and empty-target flagging, legacy passthrough with no context, floor to triggering event / to last visible message / reply downgrade / reaction drop, one-retry-then-floor and one-retry-then-accept in the step, `reply_sent` payload carries `replyToActorId`, `eventHandles` ordinal assignment with and without a triggering event.

## Why

In a 30-reply sample run the model invented readable slugs for `replyToEventId` in 12 cases and left it empty in 18, because `formatEvent` showed it no event ids at all; `intent-parser.ts` passed the string through raw and `buildReplyEvent` / `buildReactionEvent` wrote `?? ""` verbatim, committing `reply_sent` / `reaction_sent` events with dangling targets. Reactions had the identical bug on `targetEventId`.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+23)
- `pnpm build` PASS, `pnpm -r typecheck` PASS across shared/engine/server/eval; server 739 passed, engine 305 passed, shared 132 passed.
- e2e roleplay smoke: mock replies that previously committed `replyToEventId: ""` now commit against the real triggering event id via the floor.

Closes #135

- Target resolution folded into the **single** existing retry arm: one re-prompt per attempt (budget permitting), combined trigger (repetition OR unresolved target), one retry-prompt version — max 2 LLM calls per pulse instead of 3.
- Retry-accept now applies the repetition guard (`isRepeat`); a near-duplicate retry with a valid handle can no longer commit (regression test added).
- Reply payload typed as `EventPayload`, removing the `SimulationEvent["payload"]` cast.
- Engine `build-perception-packet` test now uses the shared `makeAgent`/`makeEvent` fixtures; server gains a minimal shared intent-pipeline fixture module consumed by both new test files.

Semantic notes for reviewers (unifications of previously unpinned behavior):
- A retry returning a parse fallback (or throwing) on a target-only trigger now ends in `no_op` + `llm_failure` instead of flooring the original intent — consistent with the repetition arm.
- Floor telemetry reports the last attempt's bad handle (the intent actually floored).
- Spec gaps kept as implemented, disclosed: `floorTargets` also floors to the most recent visible `message_sent`/`reply_sent` when there is no triggering event (honors "never commit an unresolved target"; can silently re-target); the new `target_resolution_floored` operator event is beyond the spec text.

### DIFF
--- a/packages/engine/src/perception/__tests__/build-perception-packet.test.ts
+++ b/packages/engine/src/perception/__tests__/build-perception-packet.test.ts
@@ -1,0 +1,100 @@
+import { describe, it, expect } from "vitest";
+import { buildPerceptionPacket } from "../build-perception-packet.js";
+import type {
+  AgentState,
+  AttentionResult,
+  CommittedEvent,
+  TranslatedEmotionalState,
+} from "@perfectman/shared";
+
+function agent(): AgentState {
+  return {
+    agentId: "agent-me",
+    simulationId: "sim-1",
+    personaId: "p1",
+    presence: "active",
+    coreMood: { valence: 0, arousal: 0.5, stability: 0.8, energy: 0.6, circumplexAngle: 0, circumplexRadius: 0.5, momentumValence: 0, momentumArousal: 0 },
+    socialEmotions: { jealousy: 0, envy: 0, humiliation: 0, pride: 0, shame: 0, affection: 0, resentment: 0, suspicion: 0, admiration: 0, contempt: 0, neediness: 0, socialAnxiety: 0, fearOfExclusion: 0, desireForStatus: 0, desireForIntimacy: 0 },
+    relationalStates: new Map(),
+    memories: [],
+    initiativeAccumulators: [],
+    lastProcessedEventId: null,
+    lastActionAt: null,
+    lastRuminationPulse: null,
+    arrivalPulse: null,
+    createdAt: 0,
+    updatedAt: 0,
+  };
+}
+
+function event(id: string, actor: string, type: CommittedEvent["type"] = "message_sent"): CommittedEvent {
+  return {
+    id,
+    simulationId: "sim-1",
+    channelId: "ch-1",
+    actorId: actor,
+    type,
+    payload: { content: `msg ${id}` },
+    createdAt: 1,
+    pulseIndex: 1,
+    sourceEventIds: [],
+    emotionalSalience: "low",
+    visibility: { visibleToAgents: [], visibleToSpectators: true, visibleToOperators: true, visibilityReason: "public" },
+  };
+}
+
+const translated: TranslatedEmotionalState = {
+  moodDescription: "",
+  socialContext: "",
+  relationalFlavors: [],
+  pressureDescriptions: [],
+  inhibitionDescriptions: [],
+};
+
+const attention: AttentionResult = {
+  noticed: true,
+  dueScore: 1,
+  reasons: [],
+  needsLLM: true,
+  triggeringReason: "direct_mention",
+};
+
+describe("buildPerceptionPacket eventHandles", () => {
+  it("assigns e1 to the triggering event, then e2.. to visible-context events in order", () => {
+    const trigger = event("evt_trigger", "agent-peer");
+    const ctx1 = event("evt_ctx_1", "agent-x");
+    const ctx2 = event("evt_ctx_2", "agent-y");
+
+    const packet = buildPerceptionPacket(
+      agent(),
+      [ctx1, ctx2, trigger],
+      trigger,
+      [],
+      attention,
+      translated,
+      [],
+    );
+
+    expect(packet.eventHandles).toEqual({
+      e1: "evt_trigger",
+      e2: "evt_ctx_1",
+      e3: "evt_ctx_2",
+    });
+    // the triggering event is not double-counted as a context event
+    expect(packet.visibleContextEvents.map((e) => e.id)).toEqual(["evt_ctx_1", "evt_ctx_2"]);
+  });
+
+  it("starts handles at e1 for the context events when there is no triggering event", () => {
+    const ctx1 = event("evt_a", "agent-x");
+    const ctx2 = event("evt_b", "agent-y");
+
+    const packet = buildPerceptionPacket(agent(), [ctx1, ctx2], null, [], attention, translated, []);
+
+    expect(packet.eventHandles).toEqual({ e1: "evt_a", e2: "evt_b" });
+  });
+
+  it("returns an empty handle map when nothing is in view", () => {
+    const packet = buildPerceptionPacket(agent(), [], null, [], attention, translated, []);
+    expect(packet.eventHandles).toEqual({});
+  });
+});
--- a/packages/engine/src/perception/__tests__/build-perception-packet.test.ts
+++ b/packages/engine/src/perception/__tests__/build-perception-packet.test.ts
@@ -1,47 +1,10 @@
 import { describe, it, expect } from "vitest";
 import { buildPerceptionPacket } from "../build-perception-packet.js";
+import { makeAgent, makeEvent } from "../../__tests__/fixtures.js";
 import type {
-  AgentState,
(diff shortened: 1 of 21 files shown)


matheusht:
**Two-axis review** (per the [code-review skill](https://github.com/mattpocock/skills/blob/main/skills/engineering/code-review/SKILL.md)): two independent agents reviewed `main...17a4e92` — **Standards** (repo standards + Fowler smell baseline) and **Spec** (issue #135). Static review; axes kept separate, no reranking.

## Standards
0 hard violations; 3 findings (1 low-severity documented-guideline breach, 2 judgement calls):

1. **Q5 (testing-strategy): hand-rolled fixtures instead of existing engine helpers** — the new engine test `build-perception-packet.test.ts:10-45` re-implements `agent()` (full `AgentState`) and `event()` (full `CommittedEvent`) although `packages/engine/src/__tests__/fixtures.ts` already exports `makeAgent` and `makeEvent(overrides)` used by sibling tests. Low severity — ids needed overriding, which `makeEvent(overrides)` supports.
2. **Duplicated Code** — `intent-parser-target-resolution.test.ts:44-67` clones `replyJson()`/`reactJson()` nearly verbatim from `action-intent-step.test.ts` and duplicates an `event()` builder from `prompt-builder.test.ts` fixtures. Judgement call: server has no per-package helper file (its convention is per-file `makeEvent`), so this extends existing practice — but these two new files are the moment to extract it.
3. **Primitive Obsession-adjacent cast** — `buildReplyEvent` (`intent-resolver.ts:209-215`) hoists `const payload: Record<string, unknown>` out of the literal, losing contextual typing and forcing `payload as SimulationEvent["payload"]`. Typing `channelVisibility`'s payload return as `EventPayload` (the existing domain type, `packages/shared/src/event/event.types.ts:30`) removes the cast.

## Spec
All 5 build bullets and all 4 acceptance criteria implemented and tested (eventHandles map, `[eN]` prefixes, parser resolution with retry + floor, downgrade/drop, `replyToActorId` stamping). Four findings:

1. **(P2) Target retry is a new parallel arm, not the existing retry arm.** Spec: "the existing action-intent retry arm re-prompts once (budget permitting)…". The PR adds a second, independent retry block in `ActionIntentStep.run` (`action-intent-step.ts:261-306`) guarded by `!fallbackApplied && !repetitionBlocked && parseResult.unresolvedTarget`, running after the pre-existing repetition-retry arm. A pulse tripping both arms makes main call + repetition retry + target retry = **3 LLM calls** with two distinct retry prompt versions. Budget guard (`llmBudget.canCall`) exists, but the spec's single-retry-arm shape did not ship.
2. **(P2) Target-retry accept skips the repetition guard.** The repetition arm accepts only via `!retryParse.fallbackApplied && !isRepeat(retryParse.intent)`; the new target-retry arm accepts with `!retryParse.fallbackApplied && !retryParse.unresolvedTarget` — `isRepeat`/`isNearRepeat` is never re-run on `retryParse.intent`. Trigger: first call returns an intent with an unresolvable handle and no repetition trip, retry returns a near-duplicate message with a valid handle → it commits, bypassing the repetition guard that governs every other accept path in this step.
3. **(P3) Floor extends to last visible message, beyond the spec's triggering-event floor.** Spec: "infer the triggering event as the target… Only when there is no triggering event and no visible message: downgrade the reply to a plain message, drop the reaction." `floorTargets` (`intent-parser.ts:296-308`) adds a second floor source: the most recent visible `message_sent`/`reply_sent`. Defensible reading, honors "never commit an empty or unresolved target", but the reply can now silently re-target an arbitrary recent message.
4. **(P3) New operator telemetry `target_resolution_floored` not in spec** (`operator.types.ts:98-107`). Consistent with existing operator-event patterns and low-risk; noted for completeness.

---
**Per axis:** Standards — 3 findings (worst: Q5 fixture duplication, P2) · Spec — 4 findings (worst: parallel retry arm up to 3 LLM calls/pulse + skipped repetition guard on retry accept, P2×2).

yoarajota:
@matheusht nice catch on the second retry arm, that was a real miss vs the spec — folded target resolution into the single retry arm, max 2 llm calls per pulse now, and the retry-accept runs the repetition guard too. added a regression test for the near-duplicate-with-valid-handle case. consolidated the new fixtures in `packages/server/src/__tests__/fixtures.ts`, the engine test uses makeAgent/makeEvent now, and the payload is typed as EventPayload so the cast is gone. on the floor reaching the last visible message: I think its defensible, it honors the never commit an unresolved target line, but it can silently re-target so I documented it in the body... target_resolution_floored stays too, follows the existing operator event pattern

caiotheodoro:
tl;dr: Good approach to model event referencing. No blockers.

- Short ordinal handles (e1, e2, ...) reduce model hallucination on raw event IDs — the model picks from a bounded set instead of inventing UUIDs.
-  as a retry-exhausted fallback (infer triggering event as reply target) is a reasonable floor — better than falling back to no_op.
-  as optional 4th param to  preserves backward compat for callers that don't need target resolution.
- The bracket-wrapped handle tolerance () is a good catch — models do emit that.

caiotheodoro (approved):
LGTM

Issue #124, opened 2026-08-30
Decide: lastActionAt write mechanism + initiative-accumulator relief model


yoarajota:
Child of the [simulation runtime & behavior fix plan](https://github.com/caiotheodoro/perfectman/issues/119) map. **Type: grilling** (HITL). Head of the causal spine — graduated from #120. Consult `docs/research/lastactionat-root-cause.md`.

## Question

How is `lastActionAt` written, and what is the initiative-accumulator relief model?

Confirmed root cause (#120): `AgentState.lastActionAt` is never written non-null, so `justActed` (`packages/engine/src/step/run-engine-step.ts:103`) is always false, the only relief path (`packages/engine/src/initiative/update-initiative-accumulators.ts:75-78`) is unreachable, and all 17 initiative sources clamp at `1.0` permanently.

## Decisions to make

1. **Write mechanism** — scheduler stamps `updatedAgentState.lastActionAt` post-commit, OR engine reads the already-correct `snapshot.rateLimitStatus.lastActionAt` (`engine-snapshot-projection.ts:44`, sourced from `rate-limit-gate.ts:62`), OR both.
2. **What counts as "acted"** — the rate-limit gate currently tracks message / reply / create_channel only and excludes reactions. Is a reaction an action for relief purposes?
3. **Relief shape** — global "decay every accumulator on any action" (current design intent) vs targeted per-motive relief; proportional decay vs reset-to-zero; whether to add a passive `else`-branch decay for agents who stay silent (none exists today).
4. **Dead code around this** — wire or delete `emotionBoost` (`updateInitiativeAccumulators` is only ever called with `zeroActionEmotions()`, so `SOURCE_EMOTION_MAP` modulation never runs); drop the unused `lastActionAt` / `pulseIntervalMs` params of `scoreInitiativeCandidates`; confirm the never-exercised attention post-action cooldown magnitude (0.20) / horizon (`score-attention.ts:154-159`).
5. `cold_start_bootstrap` staying pinned at `1.0` 40+ pulses in is a symptom of the same unreachable relief — confirm it's covered by the relief fix and needs no separate handling.

## Resolves

The write mechanism, the "acted" definition, the relief model, and the disposition of each piece of adjacent dead code.

yoarajota:
## Resolution

[...]

Issue #119, opened 2026-08-30
Perfectman simulation runtime & behavior fix plan


yoarajota:
Wayfinder map. Charted 2026-08-30 from a single `deepseek-api` simulation run (44 pulses, 2 `observer` personas, `deepseek-v4-flash`; log at `tmp/sim-run-2026-08-30.log`). The run worked end-to-end with zero LLM errors but exposed a spread of runtime defects, saturated internal state, and V1 features that produced nothing.

## Destination

A prioritized fix plan for the perfectman simulation runtime, ranked in causal order (root-cause chain first, severity as tiebreaker). The plan **is** this issue's Decisions-so-far: one line per resolved ticket, each linking the ticket that holds the detail. The map is done when nothing is left to decide and someone can run `to-tickets` off Decisions-so-far. Planning only — decisions, not diffs.

## Notes

- **Domain**: event-oriented social-presence engine. Pipeline: `event -> visibility -> attention -> interpretation -> motivation -> emotion -> pressure -> inhibition -> intent -> resolver -> committed event`. Packages: `engine` (cognition steps), `server` (runtime, delivery, pulse scheduler), `shared` (types), `eval`.
- **Skills every session should consult**: `grilling` + `domain-modeling` for any decision ticket; `research` for research tickets.
- **Scope**: engine/runtime defects, behavioral tuning, detection/observability, and verifying the "implemented but never deep-tested" V1 features (memory, relational state, narrative layer).
- **Evidence base**: the one run above. A wider validation run is queued (see #129), to land after the behavioral fixes.
- **Research findings** live in the resolution comments on #120, #121, #122, #131 (the working `docs/research/*.md` drafts were dropped once their substance was on the tickets).
- **Off-map mechanical fixes** (no decision, filed as plain issues):
  - #116 — stdout delivery gateway emits each `agent_message` twice
  - #117 — prompt assembly can exceed configured `maxInputTokens`
  - #118 — `createdAt` has two time bases across event types
  - #130 — spectator layer: `snapshot.html` writes to `packages/server/tmp/` not repo root (cwd bug), hint cap breaks mid-word, `reaction_sent` emoji dropped

## Decisions so far

[...]

PR #142, opened 2026-08-30
fix(engine): relieve initiative accumulators stuck at 1.0 (#132)

### LINKED ISSUE #132: lastActionAt write + initiative-accumulator relief
## Parent

Decision: #124 (initiative-accumulator relief model). Plan: #119.

## What to build

An agent that acts has its initiative accumulators discharged, and an agent that stays silent has them slowly relaxed — so no motive pins permanently and motive selection keeps dynamic range over a long run.

Today `AgentState.lastActionAt` is never written, so the `justActed` relief branch is unreachable and all 17 initiative sources clamp at `1.0` within ~18 pulses and stay there.

- The pulse scheduler stamps `lastActionAt = now` on the agent state it persists, after that agent commits an outward social act (message / reply / reaction / channel creation) — not for `no_op` or memory-only outcomes.
- Add a passive decay on the non-acted branch of the accumulator update (`value·(1 − decayRate·0.5)`), so a silent agent's motives also relax.
- Delete the unused `lastActionAt` / `pulseIntervalMs` parameters of `scoreInitiativeCandidates`.
- Leave `emotionBoost` / `SOURCE_EMOTION_MAP` in place with a `TODO(owner)` pointing at the "emotion-modulated accumulator growth" fog item; leave the attention post-action cooldown active.
- Document the accumulator dynamics model (growth, on-action relief, passive decay, 5-pulse cooldown) — the `0.5` passive coefficient and the cooldown magnitude are provisional, to be tuned in #129.

## Acceptance criteria

- [ ] After an agent commits a social act, its `agent_state_snapshot` shows `lastActionAt` set and its initiative accumulators lower than the prior pulse.
- [ ] Over a 40-pulse 2-agent run, no initiative source stays at `1.0` across consecutive pulses.
- [ ] A silent agent's accumulators trend down, not monotonically up.
- [ ] `scoreInitiativeCandidates` no longer takes `lastActionAt` / `pulseIntervalMs`; build + tests green.
- [ ] The accumulator dynamics model is documented alongside the code.

## Blocked by

None (can start immediately).

yoarajota:
## What

Unsticks the initiative-accumulator relief path so the 17 sources stop pinning at `1.0`:

- The pulse scheduler stamps `stepResult.updatedAgentState.lastActionAt = now` before `persistAndSnapshot` when the committed act is `send_message` / `reply_to_message` / `react` / `create_channel`. `no_op`, `write_memory`, and typing/lifecycle intents are excluded; on `fallback_committed` the `fallbackIfBlocked` type decides.
- Passive decay on the silent branch of `updateInitiativeAccumulators`: `value *= 1 - decayRate * 0.5` before growth, giving a quiet agent a fixed point well below `1.0` instead of a monotone climb. The `0.5` coefficient is provisional, retuned in #129.
- `scoreInitiativeCandidates` drops its never-read `lastActionAt` and `pulseIntervalMs` parameters; both call sites updated.
- `SOURCE_EMOTION_MAP` / `emotionBoost` kept as-is with a `TODO(owner: team)` pointing at the emotion-modulated-growth item on #119. The accumulator-dynamics model (growth, global `1 - decayRate*2` relief on `justActed`, passive decay, 5-pulse `lastFiredAt` cooldown, the provisional `score-attention` `0.20` / `pulseIntervalMs*3` post-action cooldown) is documented alongside the function.
- `hollow-completion` eval recipe gains `SIDELINE_CHANNEL` + a per-review `sidelineBurst()` ana cannot see, so its log divergence rides on explicit invisible events instead of the per-pulse operator-only no-op noise that relaxed accumulators remove.
- 13 tests: engine — silent relaxation above the fixed point, 100-pulse convergence `< 1.0`, no source pinned at `1.0` across 40 consecutive silent pulses, `scoreInitiativeCandidates` arity, `arrivalPulse` still gates; server — each of the four outward-act types stamps `lastActionAt`, and `no_op` / `write_memory` / blocked / fallback-to-`no_op` do not while fallback-to-`send_message` does.

## Why

`AgentState.lastActionAt` was never written non-null, so `justActed` in `run-engine-step` was always false, the sole relief branch in `update-initiative-accumulators` was unreachable, and every initiative source clamped at `1.0` within ~18 pulses and stayed there. Parent decision #124, plan #119.

## Validation

- `pnpm lint` per-package typecheck PASS for `shared` / `engine` / `server`, `eval` compiles clean via `pnpm build` (aggregate `pnpm -r typecheck` OOMs in this sandbox) - `pnpm test:all` PASS 1328 (+13)
- `pnpm build` PASS all 4 packages; `pnpm --filter @perfectman/engine test` 307 (+5); `pnpm --filter @perfectman/server test` 727 (+8)
- `hollow-completion` divergenceFromLog `0.34 -> 0.30` after the accumulator change dropped it under the `0.33` meaning-made ceiling and ended the run at pulse 2; the recipe fix restores the 120-pulse cap with every resolve record `kind: re_goal`, reason `hollow completion re-goals`

Closes #132

- Added `initiative-two-agent-run.test.ts`: a real 40-pulse two-agent run (canned AgentRuntime, no live LLM, fully seeded, ~40ms) asserting no initiative source — including `cold_start_bootstrap` — stays at 1.0 across consecutive pulses. Mutation-checked: it fails when both decay and relief are removed.
- AC2 coverage note: the engine-level silent-loop test skips `cold_start_bootstrap`; the new run-level test covers it.

Spec deviation, disclosed: `cold_start_bootstrap` is exempt from the passive-decay formula — its decayed fixed point would be `0.4·energy`, below the 0.30 firing threshold for energy ≤ 0.75, so the source could never fire. It remains growth-only + clamp; the run-level fire→relief cycle caps it. Residual risk: it is the only source that could still pin at 1.0 if every initiative-driven act fails to commit outward.

### DIFF
--- a/packages/engine/src/__tests__/decision.test.ts
+++ b/packages/engine/src/__tests__/decision.test.ts
@@ -189,6 +189,11 @@ describe("updateInitiativeAccumulators + scoreInitiativeCandidates", () => {
     const next = updateInitiativeAccumulators(saturated, ZERO_ACTIONS, agent, CAIO, 1, false);
 
     for (const acc of next) {
+      if (acc.source === "cold_start_bootstrap") {
+        // passive-decay exempt: growth-only, so it rises instead of relaxing
+        expect(acc.value).toBeGreaterThan(0.9);
+        continue;
+      }
       expect(acc.value).toBeLessThan(0.9);
     }
   });
@@ -199,7 +204,10 @@ describe("updateInitiativeAccumulators + scoreInitiativeCandidates", () => {
     for (let i = 0; i < 100; i++) {
       accumulators = updateInitiativeAccumulators(accumulators, ZERO_ACTIONS, agent, CAIO, i, false);
     }
+    // cold_start_bootstrap is passive-decay exempt (its 0.30 threshold sits
+    // below the 0.4*energy ceiling); every other source relaxes below 1.0.
     for (const acc of accumulators) {
+      if (acc.source === "cold_start_bootstrap") continue;
       expect(acc.value).toBeLessThan(1);
     }
     // reply_pressure is the fastest grower; its fixed point is
@@ -217,11 +225,29 @@ describe("updateInitiativeAccumulators + scoreInitiativeCandidates", () => {
       const current = new Map(accumulators.map(a => [a.source, a.value]));
       if (prev) {
         for (const [source, value] of current) {
+          if (source === "cold_start_bootstrap") continue;
           const stuckAtCeiling = value >= 1 && prev.get(source)! >= 1;
           expect(stuckAtCeiling, `${source} pinned at 1.0 on pulse ${i}`).toBe(false);
         }
       }
       prev = current;
+    }
+  });
+
+  it("cold_start_bootstrap still crosses its threshold within ~2 silent pulses for a calm agent", () => {
+    // Passive-decay exemption regression: with universal passive decay,
+    // cold_start_bootstrap's fixed point (0.4*energy) sits under its 0.30
+    // threshold for typical persona energy, so proceed would never fire.
+    for (const energy of [0.3, 0.5]) {
+      const agent = makeAgent("cold", { coreMood: { ...BASE_MOOD, energy } });
+      let accumulators: InitiativeAccumulator[] = [];
+      let proceeded = false;
+      for (let i = 0; i < 2; i++) {
+        accumulators = updateInitiativeAccumulators(accumulators, ZERO_ACTIONS, agent, CAIO, i, false);
+        const candidates = scoreInitiativeCandidates(accumulators, i, null);
+        if (candidates.find(c => c.source === "cold_start_bootstrap")!.proceed) proceeded = true;
+      }
+      expect(proceeded, `energy ${energy}`).toBe(true);
     }
   });
 
@@ -250,11 +276,8 @@ describe("updateInitiativeAccumulators + scoreInitiativeCandidates", () => {
     expect(bootstrap!.score).toBeGreaterThan(0);
   });
 
-  it("scoreInitiativeCandidates no longer accepts lastActionAt / pulseIntervalMs", () => {
+  it("scoreInitiativeCandidates takes 3 args and still gates on arrivalPulse as the third", () => {
     expect(scoreInitiativeCandidates.length).toBe(3);
-  });
-
-  it("scoreInitiativeCandidates still gates on arrivalPulse as its third argument", () => {
     const accumulators = updateInitiativeAccumulators([], ZERO_ACTIONS, makeAgent(), CAIO, 0, false)
       .map(a => ({ ...a, value: 0.99 }));
     const beforeArrival = scoreInitiativeCandidates(accumulators, 2, 5);
--- a/packages/engine/src/__tests__/decision.test.ts
+++ b/packages/engine/src/__tests__/decision.test.ts
@@ -180,6 +180,51 @@ describe("updateInitiativeAccumulators + scoreInitiativeCandidates", () => {
     expect(afterAct).toBeLessThan(beforeAct);
   });
 
+  it("passive decay: a silent pulse relaxes every accumulator that is above its fixed point", () => {
+    const agent = makeAgent();
+    const saturated: InitiativeAccumulator[] = (
+      updateInitiativeAccumulators([], ZERO_ACTIONS, agent, CAIO, 0, false)
+    ).map(a => ({ ...a, value: 0.9 }));
+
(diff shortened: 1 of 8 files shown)


matheusht:
**Two-axis review** (per the [code-review skill](https://github.com/mattpocock/skills/blob/main/skills/engineering/code-review/SKILL.md)): two independent agents reviewed `main...8d71492` — **Standards** (repo standards + Fowler smell baseline) and **Spec** (issue #132). Static review; axes kept separate, no reranking.

## Standards
0 hard violations; 1 judgement call:

- **Primitive Obsession / Repeated Switches-adjacent** — `OUTWARD_SOCIAL_ACT_TYPES = new Set(["send_message", "reply_to_message", "react", "create_channel"])` (`pulse-scheduler.ts:40-44`) expresses the "outward social act" domain concept as a bare string set in server, while the engine already curates near-identical intent-type groupings (`rate-limit-rules.ts` `MESSAGE_INTENT_TYPES`, `compute-available-actions.ts` `ALL_INTENT_TYPES`). Future intent additions must remember to update this set. Judgement call — the master-contract ownership map gives server the stateful orchestration and engine the pure rules, and the constant is documented, so the repo standard arguably overrides the smell.

## Spec
Nearly every line of issue #132 implemented: scheduler stamping of `lastActionAt` for the four outward act types (with `no_op`/`write_memory`/blocked/fallback handling tested), passive decay on the silent branch, deletion of the unused `scoreInitiativeCandidates` params with call sites migrated, `TODO(owner)` on `SOURCE_EMOTION_MAP`, attention cooldown left active, dynamics model documented. Three findings:

1. **(P2) cold_start_bootstrap exempted from the spec'd passive-decay formula.** Spec: "Add a passive decay on the non-acted branch of the accumulator update (value·(1 − decayRate·0.5)), so a silent agent's motives also relax." The implementation keeps `cold_start_bootstrap` growth-only on the silent branch (`update-initiative-accumulators.ts:133-145`). The deviation is mathematically forced — its decayed fixed point would be `0.4·energy`, below the 0.30 threshold for energy ≤ 0.75, so the source could never fire (proven by the new regression test) — but it contradicts the spec line and leaves the one source that can still pin at 1.0 (growth-only + clamp; reachable if every initiative-driven act fails to commit outward), the exact failure mode AC2 targets, whose test now skips this source. Deliberate and documented in code and commits 5c815896/778ca0fc, but absent from the PR body's "What" section and unreconciled with the issue text.
2. **(P2) AC2 "2-agent run" only tested as a single-agent loop, one source excluded.** Spec AC: "Over a 40-pulse 2-agent run, no initiative source stays at 1.0 across consecutive pulses." The added engine test (`decision.test.ts:45-72`) is a single-agent pure-function loop, not a 2-agent run, and explicitly `continue`s on `cold_start_bootstrap` — so the 2-agent aspect is untested and 1 of 17 sources is excluded from the pin check.
3. **(P3) hollow-completion eval recipe rework is outside issue scope.** Issue #132's "What to build" has five bullets, none mentioning eval scenarios. The diff adds `SIDELINE_CHANNEL` and a per-review `sidelineBurst()` to `packages/eval/src/goal-layer/scenario-recipes.ts:388-411` (commit 8c878d31). The PR body justifies it — accumulator relaxation dropped `hollow-completion` divergenceFromLog from 0.34 to 0.30, under the 0.33 meaning-made ceiling, ending the run at pulse 2 — so it is a disclosed, necessary consequence-repair, but beyond the issue's stated scope.

---
**Per axis:** Standards — 1 finding (outward-act string set, P3) · Spec — 3 findings (worst: cold_start_bootstrap decay exemption leaves one source able to pin at 1.0 and untested against AC2, P2×2).

yoarajota:
@matheusht added `initiative-two-agent-run.test.ts` — real 40-pulse run with a canned runtime, no llm, covers cold_start_bootstrap too, and I mutation checked it: remove decay and relief together and it fails on the pin. on the decay exemption: oh yeah it deviates from the spec line but the math forces it, the decayed fixed point would be 0.4*energy and proceed requires > 0.30, so for energy <= 0.75 the source would never fire... disclosed in the body. the string set and the eval recipes I'm keeping — one is documented and the ownership map puts it in server, the other is a disclosed consequence-repair

caiotheodoro:
tl;dr: Passive decay fix is well-motivated. No blockers.

- The fixed-point math () explains why values converge below 1.0 instead of re-saturating.  exemption is justified — its threshold sits below the passive-decay ceiling.
- Removing  and  from  is correct — both were dead parameters after the attention cooldown moved to .
- The 40-pulse silent-run test catches the exact regression this fixes (any source pinned at 1.0 across consecutive pulses).
- The two-agent run test in server validates the fix end-to-end through the pulse scheduler.

caiotheodoro (approved):
LGTM

PR #144, opened 2026-08-30
fix(server): emit one agent_message per committed message (#116)

### LINKED ISSUE #116: stdout delivery gateway emits each agent_message twice
Mechanical fix surfaced by the 2026-08-30 `deepseek-api` simulation run (`tmp/sim-run-2026-08-30.log`). Filed off the [simulation runtime & behavior fix plan] map — no design decision, obvious fix.

## Symptom

Every `agent_message` is written to stdout **twice**. The pairs are byte-identical and adjacent:

```
line  8: {"type":"agent_message",...,"agentId":"ana","content":"oi gente"...}
line  9: {"type":"agent_message",...,"agentId":"ana","content":"oi gente"...}
line 17/18: bruno "oi ana"  (identical pair)
line 32/33: ana "tudo bem?" (identical pair)
```

122 `agent_message` events for 59 real messages (44 `send_message` + 15 `reply_to_message` intents).

## Where

Single write site: `packages/server/src/delivery/stdout-delivery-gateway.ts:18` — `this.write("agent_message", { channelId, message })`.

## Hypothesis

Delivery fans out per channel member (2 members in the run) and logs a per-channel line on each pass, instead of once per committed message. Confirm whether the gateway's `deliver` is invoked per-recipient and dedupe to one emission per message.

## Acceptance

One `agent_message` line per committed message on the stdout gateway.

yoarajota:
## What

Collapses the per-member delivery fan-out to a single emission per committed message:

- `DeliveryProjection.project` looped `for (const agentId of memberAgentIds)` and called the channel-scoped `sendAgentMessage(channelId, msg)` once per member, with a byte-identical `{ channelId, message }` payload each pass — an N-member channel produced N calls on every gateway (`StdoutDeliveryGateway`, `DiscordDeliveryGateway`, `CompositeDeliveryGateway`).
- `message_sent`, `reply_sent`, `reaction_sent` now emit once, gated by a new `visibleToAnyMember` helper that preserves the existing `filterVisibleEventsForAgent` semantics (restricted `visibleToAgents`, private-channel non-members).
- Fixed at the caller, not `StdoutDeliveryGateway`: the double-invoke hit every gateway, so the stdout-only symptom was one facet of a caller bug — the stdout gateway's single `this.write("agent_message", ...)` was already correct.
- `3 tests`: message to a 4-member channel emits one `sendAgentMessage`; reply + reaction each emit once on an N-member channel; `StdoutDeliveryGateway` writes exactly one `agent_message` line for a 4-member channel. Existing throwing-gateway assertion retightened from per-member `operatorEvents` count to `1`.

## Why

Every `agent_message` was written to stdout twice, pairs byte-identical and adjacent — 122 lines for 59 real messages in a 2-member run. The projection was treating a channel-scoped gateway call as if it were per-recipient.

## Validation

- `pnpm -r typecheck` PASS (per package) - `pnpm test:all` PASS (`1318 passed`, +3); root `pnpm lint` aggregator OOMs on concurrency, unrelated to this change.
- `pnpm --filter @perfectman/server test` — `722 passed`; new end-to-end test spies `process.stdout.write` and asserts exactly 1 `agent_message` line for a 4-member channel.

Closes #116

### DIFF
--- a/packages/server/src/simulation/__tests__/delivery-projection.test.ts
+++ b/packages/server/src/simulation/__tests__/delivery-projection.test.ts
@@ -173,16 +173,27 @@ describe("DeliveryProjection — one emission per committed message", () => {
   const AGENT_3 = "agent_3";
   const AGENT_4 = "agent_4";
   const BIG_CHANNEL_ID = "ch_big";
+  const NOW = 1_700_000_000_000;
+
+  let gateway: MockDeliveryGateway;
+  let projection: DeliveryProjection;
+
+  beforeEach(() => {
+    gateway = new MockDeliveryGateway();
+    projection = new DeliveryProjection(gateway);
+  });
 
   const bigChannel: Channel = {
     ...PUBLIC_CHANNEL,
     id: BIG_CHANNEL_ID,
     memberAgentIds: [AGENT_1, AGENT_2, AGENT_3, AGENT_4],
+    createdAt: NOW,
+    updatedAt: NOW,
   };
   const bigMembership: ChannelMembership[] = [AGENT_1, AGENT_2, AGENT_3, AGENT_4].map(agentId => ({
     channelId: BIG_CHANNEL_ID,
     agentId,
-    joinedAt: Date.now(),
+    joinedAt: NOW,
   }));
   const bigChannelEvent = (type: CommittedEvent["type"]): CommittedEvent =>
     makeCommittedEvent({
@@ -192,15 +203,11 @@ describe("DeliveryProjection — one emission per committed message", () => {
     });
 
   it("a message to a 4-member channel produces exactly one sendAgentMessage call", async () => {
-    const gateway = new MockDeliveryGateway();
-    const projection = new DeliveryProjection(gateway);
     await projection.project(bigChannelEvent("message_sent"), [bigChannel], bigMembership, SETTINGS);
     expect(gateway.agentMessages).toHaveLength(1);
   });
 
   it("reply and reaction events each emit exactly once on an N-member channel", async () => {
-    const gateway = new MockDeliveryGateway();
-    const projection = new DeliveryProjection(gateway);
     await projection.project(bigChannelEvent("reply_sent"), [bigChannel], bigMembership, SETTINGS);
     await projection.project(bigChannelEvent("reaction_sent"), [bigChannel], bigMembership, SETTINGS);
     expect(gateway.agentMessages).toHaveLength(2);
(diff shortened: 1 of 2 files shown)


matheusht:
**Two-axis review** (per the [code-review skill](https://github.com/mattpocock/skills/blob/main/skills/engineering/code-review/SKILL.md)): two independent agents reviewed `main...d158d93` — **Standards** (repo standards + Fowler smell baseline) and **Spec** (issue #116). Static review; axes kept separate, no reranking.

## Standards
0 hard violations; 3 judgement calls:

1. **Duplicated Code (Q4/Q5)** — the three new tests in `delivery-projection.test.ts:186-212` each repeat `const gateway = new MockDeliveryGateway(); const projection = new DeliveryProjection(gateway);` inline, while the sibling describe shares this arrangement via `beforeEach`. testing-strategy Q5 ("recurring arrangements come from shared per-package test helpers") points to a local `beforeEach`.
2. **Duplicated Code / Repeated Switches** — `delivery-projection.ts:24-66`: the patch adds an identical `visibleToAnyMember` guard to all three event cases and leaves the pre-existing triplicated guard → build-msg → `safeGatewayCall` shape in place; a private helper taking the partial `DeliveryMessage` would collapse it. Pre-existing structure, and the new guard itself already routes through the extracted helper — hence judgement call only.
3. **Q7 determinism** — new fixtures `bigChannel`/`bigMembership` call `Date.now()` for `joinedAt`/`createdAt`. Mirrors the file's pre-existing fixture convention and cannot affect the asserted invariant (emission count), but new fixture code could inject a fixed timestamp.

Conformance notes: one concern, server projection layer only per the master-contract ownership map ("Delivery implementations must not compute visibility" — visibility correctly stays in `DeliveryProjection` via `filterVisibleEventsForAgent`, no gateway file touched); why-explained in the body; every `!` follows a prior existence proof; the stdout spy is asserted and restored in `finally`; the `JSON.parse(line) as { type: string }` cast is narrow, not a banned `as any`.

## Spec
No missing requirements, no scope creep, no implemented-but-wrong behaviour. The per-member fan-out loop was collapsed to a single emission per committed message, previous visibility semantics preserved exactly via `visibleToAnyMember`, and the acceptance criterion (one `agent_message` line per committed message on stdout) is proven by a new end-to-end stdout-spy test on a 4-member channel. Fixing at the caller rather than `StdoutDeliveryGateway` matches the issue's root-cause framing (the bug hit every gateway).

---
**Per axis:** Standards — 3 findings (worst: triplicated guard→build→send shape, P3) · Spec — 0 findings.

yoarajota:
@matheusht nice catch. shared the arrangement in a beforeEach, extracted `deliverIfVisible` so the guard-build-send shape lives in one place, and pinned the new fixtures to fixed timestamps

caiotheodoro:
tl;dr: Correct fix. No blockers.

- The old per-member loop was semantically wrong —  targets a channelId, not individual agents, so per-member visibility filtering was redundant overhead.
-  +  is the right abstraction. The N-member channel tests (4 agents, stdout gateway) confirm exactly one emission.
-  is preserved unchanged — error handling untouched.

caiotheodoro (approved):
LGTM

Issue #138, opened 2026-08-30
Relational-state accretion: emit participant identifiers, message_sent rule, isTarget fix, real-resolver test


yoarajota:
## Parent

Decision: #127 (relational-state accretion). Plan: #119.

## What to build

A two-agent conversation actually moves relational state — the accretion machinery that runs every pulse but currently no-ops starts producing entries.

- The intent resolver's message / reply / reaction event builders write participant identifiers into the event payload: `personTargets` (from the intent), `mentionedAgentIds` (from a minimal mention parser over the visible content — exact display-name / `@handle` match), and `replyToActorId` (from slice #6's resolution).
- Add a `message_sent` rule to the relational update rules with small deltas (a plain message is less loaded than a reply or reaction); with no `personTargets`, a small co-presence bump for each channel member.
- Fix relational target detection so it also consults `invitedAgentIds` (so a channel invite accretes an entry for the invitee).
- A regression test drives an intent through the real intent resolver, feeds the resulting committed event to relational accretion, and asserts a non-empty `relationalStates`.

## Acceptance criteria

- [ ] After a short 2-agent conversation, both agents have a non-empty `relationalStates` entry for each other.
- [ ] A reply produces a larger relational delta than a plain message; a reaction registers.
- [ ] Creating a channel and inviting an agent accretes a relational entry for that invitee.
- [ ] The regression test exercises the real event builders, not hand-injected payloads.

## Blocked by

#135 (reply/reaction target resolution — provides replyToActorId).

yoarajota:
Delivered via #151 (participant identifiers on message/reply/reaction payloads, message_sent relational rules, isTarget invitee fix, real-resolver regression test), merged to main. Follow-ups tracked in #129 (resentment-floor re-tune) and #152 (eval-harness projection gap).

Issue #136, opened 2026-08-30
Memory: formation via prompt + persistence projection into agentState.memories


yoarajota:
## Parent

Decision: #126 (memory formation trigger + persistence path). Plan: #119.

## What to build

An agent forms a first-person memory during a normal turn, and that memory persists into its state and shapes its later prompts.

- The action-intent prompt instructs the model, with criteria, when to emit `memoryWrites` on a normal action intent (it currently lists the field but never asks for it).
- A pulse-scheduler projection consumes committed `memory_written` events and appends them to the acting agent's `agentState.memories`, persisted like any other projected state.
- Remove `write_memory` from the available-actions list — memory is a side-channel on a normal action, not a turn-consuming choice. (Leave the schema / resolver plumbing.)

## Acceptance criteria

- [ ] In a run, a `memory_written` event is produced from an agent's action intent.
- [ ] The next `agent_state_snapshot` for that agent shows the memory in `memories`.
- [ ] The memory appears in that agent's next action-intent prompt ("What you remember").
- [ ] `write_memory` no longer appears in any agent's permitted actions.

## Blocked by

#134 (intensity schema field).

yoarajota:
Delivered via #150 (memory formation via prompt + `applyMemoryProjection` persistence into agentState.memories), merged to main in 8f54305 stack.

PR #151, opened 2026-09-02
feat(engine): accrete relational state from chat events

yoarajota:
## What

Makes a two-agent conversation actually move relational state (implements #138):

- **Payload identifiers**: message/reply/reaction builders stamp `personTargets` (from the intent), `mentionedAgentIds` (minimal display-name/`@handle` mention parser over visible content) and `replyToActorId` (#135's resolution) into the payload
- **`message_sent` rule**: new entry in `RELATIONAL_UPDATE_RULES` — reply/reaction deltas > plain message; with no `personTargets`, a small co-presence bump reaches the members the visibility filter already bounds (data-only, no `channelMemberIds` stamping, no engine signature change)
- **Invitee targeting**: `isTarget` consults `invitedAgentIds` and folds the singular `invitedAgentId`, so create-and-invite accretes an entry for the invitee
- **Mentions wiring**: resolver takes an optional `agentNames` (agentId -> display name) map; the scheduler populates it from the configured personas — an unwired map degrades to zero mentions parsed, same as before
- **Regression test**: drives an intent through the real `IntentResolver`, feeds the committed event to relational accretion, asserts a non-empty `relationalStates` — no hand-injected payloads
- `html-receiver-parity` e2e: dormant-path activation keeps every canned agent attentive, so the `staleFrames > 0` fixture-adequacy counter is removed (staleness-axis coverage tracked as a follow-up)

## Why

Same evidence run as #136: `relationalStates: {}` stayed empty for both agents across a 3-message exchange — relationship pressure never built, starving the initiative sources that key off it.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS
- 1423 tests / 126 files on this branch; PR gate locally replicated and green: `BENCH GATE PASSED` — 35 golden runs, signal pass 100.0% after recalibrating the three resentment floors the relational pathway shifts (`stagnation-attractors.ts` 0.5->0.42, 0.3->0.24; `v1-behaviors.ts` 0.12->0.10), measured values recorded inline, re-tune in #129
- eval canary committed under `docs/eval/evidence/canary-136-138/` with a pre-wave control run attributing the shift to #138; the eval harness gap (scenario-runner bypasses the memory projection + identifier stamping) is recorded as follow-up #152
- Stacks on #150 (#136) — the shared `pulse-scheduler.ts` regions are disjoint; rebases after #150 merges

### DIFF
--- a/docs/adr/0013-memory-formation-and-relational-accretion.md
+++ b/docs/adr/0013-memory-formation-and-relational-accretion.md
@@ -1,0 +1,79 @@
+# ADR-0013: Memory Formation And Relational Accretion
+
+**Status**: Accepted (LOCKED, 2026-09-02) — rationale and rejected alternatives inline in this ADR (decisions **D-37..D-45**).
+
+**Numbering note**: written against the merged-tree ADRs 0001..0012 (0012 duplicated: `0012-goal-layer-meaning-made-gate.md` and `0012-goal-registry-persistence.md`); in-tree decision high-water mark **D-36** (ADR-0012). The source run's decision log numbered its decisions run-locally D-1..D-10; D-1..D-9 map to **D-37..D-45** here. Run-local D-10 (two-PR merge sequencing) is not converted — sequencing decisions expire at merge. The D-37..D-45 range is this ADR's; if another unmerged slice claims the range first, renumber here and reconcile the inline D-refs.
+
+## Context
+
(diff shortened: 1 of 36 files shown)


yoarajota:
follow-ups from the run review, both closed:

- staleness axis owned again: the partial-frame unit fixture now asserts bruno's agentThinking row is absent on pulse 1 (html-snapshot-gateway.test.ts), so a receiver leaking stale rows fails at the lowest owning layer — the e2e title's promise is witnessed somewhere now
- agentNames wire pinned end-to-end: two-agent run through the real resolver, display-name mention stamps `mentionedAgentIds: ["agent_2"]`; substring exclusion ("Other Agentrix" → no mention) asserted too — the parser matches its docstring

PR #153, opened 2026-09-03
fix(server): stamp LLM rate windows with wall-clock time (#147)

### LINKED ISSUE #147: LLMBudget rate windows are inert: sim-relative LLMUsage.createdAt pruned against a wall-clock window
Found during adversarial review of #143 (the #118 `createdAt` time-base fix). Not introduced by #143 — pre-existing on `main` — and deliberately left out of that PR's scope because fixing it safely is coupled to eval determinism (see below).

## Bug

`LLMBudget.recordUsage` stores `timestamp = usage.createdAt || Date.now()` (`packages/server/src/llm/llm-budget.ts:88`). `LLMUsage.createdAt` is sim-relative ms (range ~0–10^6, sourced from `AgentRuntimeContext.now` = the pulse scheduler's sim clock). Every subsequent `cleanup()` (`llm-budget.ts:151`, called from `canCall` / `getPriority` / `getStatus`) does:

```
this.callTimestamps = this.callTimestamps.filter(t => t >= Date.now() - 60_000)
```

`Date.now() - 60_000` is ~1.7e12; every stored `timestamp` is ~10^3–10^6, so **every usage record is dropped on the next call**. Result:

- `llmCallBudgetPerMinute` never accumulates → the per-minute call cap never trips.
- `tokenBudgetPerHour` never accumulates → the per-hour token cap never trips.

Production LLM rate limiting is effectively disabled.

## Why it wasn't fixed in #143

#143 keeps `LLMUsage.createdAt` on the sim clock on purpose: it is not part of the committed/operator event stream (nothing emits it; `llmBudget.recordUsage` is its only consumer), and `AgentRuntimeContext.now` must stay deterministic for eval transcript replay + RNG seeding.

The first (reverted) attempt in #143 — `c26151d`, which passed `now: Date.now()` into the runtime context — would have *incidentally* fixed this. That same change broke `packages/eval/src/test/sweep-temperature.test.ts` ("replays an identical transcript across runs of the same scenario"): real budget accumulation plus wall-clock window-boundary crossings make otherwise-identical replays diverge. So this latent bug is currently **load-bearing for eval determinism**.

## Constraint on any fix

Must not reintroduce nondeterminism into `sweep-temperature`'s pinned-seed replay. Candidate approaches:

1. Store sim time in `callTimestamps` / token records **and** prune against sim time — but then `perMinute` / `perHour` need a sim-time-duration equivalent (pulses × `pulseIntervalMs`), which changes the semantics of those config knobs.
2. Thread a separate explicit wall-clock stamp into `recordUsage` (distinct from the determinism-carrying `ctx.now`) and keep pruning on wall-clock; ensure it never feeds anything the replay fingerprint covers.
3. Make the eval harness seed/freeze a clock so option 2 stays deterministic under test.

## Locations

- `packages/server/src/llm/llm-budget.ts:88` — `recordUsage` timestamp source
- `packages/server/src/llm/llm-budget.ts:151` — `cleanup` wall-clock window filter
- `packages/eval/src/test/sweep-temperature.test.ts` — the determinism test that constrains the fix

Context: part of the #119 simulation runtime fix plan's neighbourhood, but decision-unblocked and independent.

Elesiann:
Closes #147.

## Bug
`LLMBudgetTracker.recordUsage` stored `usage.createdAt` as the rate-window timestamp, but production callers stamp `createdAt` with the pulse-relative sim clock (~10^3-10^6). `cleanup()` prunes against `Date.now()` (~1.7e12), so every record was dropped on the next call — both rate limits never tripped.

## Fix (issue option 2)
- `recordUsage` now stamps windows with `Date.now()` taken inside the tracker; `LLMUsage.createdAt` stays on the sim clock for deterministic replay attribution and never feeds anything the replay fingerprint covers.
- `buildConfiguredSimulation` resets the sim's windows on every build, so reruns and re-attaches sharing one process (CLI, back-to-back eval runs, restart tests) don't inherit another run's usage. Limits are kept.
- Tests: shared `makeUsage` helper, 2 regression tests (sim-clock calls and tokens trip the windows), 1 expiry test (records drop out past the window, via fake timers).

## Verification
- `llm-budget` 9/9 (new tests failed pre-fix); `sweep-temperature` unchanged and green (6/6, incl. pinned-seed replay).
- `agent-runtime` (12) + `goal-layer-llm` (10) green; `tsc` clean in server + eval; `audit-tests --check` and `docs-lint --check` PASS.
- Full server suite: 717 pass; 91 failures are pre-existing environmental (missing better-sqlite3 native binding — verified identical on clean tree).
- Adversarial review (3 axes) done; findings addressed except the open point below.

## Open discussion — rate limits now really bite
Heads-up: this is a behavior flip, not just a bugfix. Defaults are 20 calls/min per simulation, and every pulse records 1 call per agent (plus retries, plus goal calls); mock counts like real. A 4-agent sim at 3s pulses does about 80 calls/min wall-clock, so most late-window calls would now be blocked as `no_op` plus `llm_budget_exceeded` where previously nothing ever blocked. Options: (a) raise the defaults, (b) scope windows per-agent instead of per-simulation, (c) accept current blocking as the intended backpressure. Team call — not changing defaults in this PR without agreement.

### DIFF
--- a/packages/eval/src/run/scenario-runner.ts
+++ b/packages/eval/src/run/scenario-runner.ts
@@ -27,6 +27,7 @@ import {
   AgentConfigRegistry,
   MockDeliveryGateway,
   createLLMProvider,
+  llmBudget,
   personaPackToProfile,
 } from "@perfectman/server";
 import type { PersonaPromptProfile } from "@perfectman/server";
@@ -146,6 +147,11 @@ class TrackingRuntime {
 
 export class ScenarioRunner {
   static async run(scenario: RoleplayScenario, opts: RunnerOpts = {}): Promise<ScenarioRunArtifact> {
+    // The cognition path records usage into the shared llmBudget singleton
+    // with wall-clock stamps (#147). Reset this scenario's windows up front
+    // so back-to-back runs in one process (e.g. the pinned-seed replay test)
+    // start from identical budget state and stay deterministic.
+    llmBudget.reset(scenario.id);
     const started = Date.now();
     const packs = new Map<string, import("@perfectman/shared").PersonaPack>();
     for (const p of ALL_PERSONAS) {
--- a/packages/server/src/config/simulation-config.ts
+++ b/packages/server/src/config/simulation-config.ts
@@ -441,6 +441,12 @@ export async function buildConfiguredSimulation(
   if (!options.llmBudget) {
     llmBudget.registerLimits(simulationId, config.simulation.settings);
   }
+  // Fresh simulation, fresh rate windows (#147): the shared singleton keeps
+  // usage keyed by simulationId across builds in one process (CLI reruns,
+  // back-to-back eval runs, restart tests re-attaching the same id), so a
+  // new build must not inherit another run's wall-clock usage. Limits are
+  // kept — reset() only clears recorded calls/tokens.
+  (budget as LLMBudgetTracker).reset(simulationId);
 
   const runtime = new SimulationRuntime({
     delivery,
--- a/packages/server/src/llm/__tests__/llm-budget.test.ts
+++ b/packages/server/src/llm/__tests__/llm-budget.test.ts
@@ -211,8 +211,58 @@ describe("LLMBudgetTracker", () => {
     expect(budget.getPriority("sim-test", "goulart")).toBe("blocked");
   });
 
-  it("should get complete status", () => {
+  it("accumulates sim-clock usage in the per-minute window (#147)", () => {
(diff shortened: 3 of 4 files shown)


yoarajota:
@Elesiann c!

yoarajota:
the ideia isnt to have quick simulations, but long running ones... we can do some A/B test to check quality... but for now we can leave as is, you current change already fix the issue.

Issue #131, opened 2026-08-30
Research: human-memory model (decay, forgetting, interference, distortion) for the memory subsystem


yoarajota:
Child of the [simulation runtime & behavior fix plan](https://github.com/caiotheodoro/perfectman/issues/119) map. **Type: research** (AFK — literature research, resolve with a subagent). **Blocks #126** (its Q5: memory decay / eviction / distortion).

## Question

What should the memory subsystem's decay, forgetting, interference, and distortion behaviour look like, grounded in the cognitive-science literature on human memory — expressed as a concrete, implementable model spec?

The retrieval side is already built (`packages/engine/src/perception/memory-salience.ts`: `scoreMemorySalience` weights relevance / memory-type / `confidence` / `unresolved` / recency; `selectRelevantMemories` caps the surfaced set). What's missing: the store (`agentState.memories`) is unbounded and never ages. #126 decided formation (prompt-driven `memoryWrites`) + persistence (a scheduler projection into `agentState.memories`); this ticket decides how those memories then behave over time.

## The existing `Memory` shape (design around this)

`packages/shared/src/memory/memory.types.ts` — fields: `type` (`pending_intention` | `relationship` | `self` | `social_theory` | `emotional_residue` | `episodic`), `summary`, `emotionalTone` (free-text string), `confidence` (0–1), `unresolved` (bool), `subjectAgentIds`, `createdAt`.

## What to produce (in `docs/research/human-memory-model.md`)

Ground each recommendation in primary/authoritative sources (cite them — Ebbinghaus forgetting curve and modern re-analyses, decay vs. interference theory, consolidation/systems consolidation, reconsolidation & memory distortion, flashbulb/emotional-memory salience, spacing/testing effects, the misinformation effect). For each, give the mechanism AND concrete parameters/curve shape keyed to the fields above:

1. **Decay** — what fades and how: `confidence` decay curve (exponential? power-law, per the forgetting-curve literature?), per how much sim time / how many pulses. Should decay rate depend on `type` and `emotionalTone` (emotional memories decay slower)?
2. **Store bound & eviction** — is there a capacity limit? If so, is eviction by lowest salience, by age, by a retrieval-strength model (e.g. ACT-R base-level activation)? Or is "eviction" really just decay-to-irrelevance with a floor?
3. **Interference** — should a new memory about the same `subjectAgentIds` / topic weaken older similar ones (retroactive interference)?
[...]

PR #157, opened 2026-09-04
feat(engine): power-law memory decay, eviction, reinforcement (#137)

### LINKED ISSUE #126: Decide: memory formation trigger + persistence path into agentState.memories
**Blocked by #131** (human-memory model — its Q5).

Child of the [simulation runtime & behavior fix plan](https://github.com/caiotheodoro/perfectman/issues/119) map. **Type: grilling** (HITL). Graduated from #121 (memory = "unbuilt in practice"). Consult `docs/research/memory-relational-state.md`.

## Question

How is agent memory formed and persisted?

No reachable path exists today: engine `memoryProposals` is hardcoded `[]` (`packages/engine/src/step/run-engine-step.ts:332`); the `memoryWrites[]` intent field (`intent-packet.schema.ts:23`) is never prompted; the `write_memory` intent commits nothing (`intent-resolver.ts:542,670-675`); and **nothing writes into `agentState.memories`** — only config init and the SQLite serializer touch it.

## Decisions to make

1. **Formation trigger** — pick one (or a combination):
   - prompt-driven inline writes (instruct the model to emit `memoryWrites` in the action intent; needs prompt + parser + resolver wiring)
   - engine salience-gated proposals (add a real memory stage to `runEngineStep`; gate on emotion magnitude / event salience)
   - scheduled background reflection (add to the surface registry `agent/surface/index.ts`, schedule it)
2. **The consumer** — build the path that moves a committed memory write into `agentState.memories` so it appears in snapshots and feeds later prompts. This is required regardless of trigger choice.
3. **`write_memory` action** — fix it to commit a `memory_written` event, or delete it as dead.
4. **Bias / decay** — the README calls for "biased, emotionally-colored memory". In scope here, or a follow-up?

## Resolves

The formation trigger, the persistence path, and the disposition of the `write_memory` action.

### LINKED ISSUE #137: Memory: power-law decay, emotional protection, eviction, recall reinforcement
## Parent

Decision: #126 Q5 + research #131 (human-memory model). Plan: #119.

## What to build

Stored memories fade over time on a power-law curve, emotionally intense ones fade slower, stale ones are evicted, and being recalled strengthens a memory.

- Fold a power-law `effectiveConfidence(memory, now)` into the memory-salience score: `confidence · (0.10 + 0.90·(1 + agePulses)^(−d_eff))`, with per-memory-type base rate `d` and emotional protection `d_eff = d·(1 − 0.6·intensity)`. Reduce the standalone recency weight (decay now carries recency).
- Eviction pass in the memory projection: drop a memory when `effectiveConfidence < 0.05` and age > 20 pulses and it is not `unresolved` / `pending_intention`; a `MAX_MEMORIES = 500` per-agent backstop evicts the lowest-strength memory on overflow. No hard capacity cap otherwise.
- On a memory being surfaced by memory selection, the scheduler emits a `memory_reinforced` operator event; the projection bumps that memory's `lastReinforcedAt`.
- The six per-type `d` values and the `0.6` / `0.05` / `20` / `500` constants are provisional — surface them as named constants for #129 tuning.

## Acceptance criteria

- [ ] A memory's effective salience decreases across pulses with no interaction.
- [ ] A high-`intensity` memory decays measurably slower than a low-`intensity` one of the same type and age.
- [ ] A low-confidence, aged (>20 pulses), non-unresolved memory is evicted; an `unresolved` one of the same age/confidence is kept.
- [ ] Surfacing a memory in a prompt raises its subsequent effective salience relative to not surfacing it.
- [ ] Decay/eviction constants are named and centralized.

## Blocked by

#136 (memory formation + persistence).

yoarajota:
## What

Adds power-law decay, eviction, and recall reinforcement to agent memory:

- `effectiveConfidence(memory, now, pulseIntervalMs)` in `memory-salience.ts`: retention `0.10 + 0.90 * (1 + agePulses)^(-d_eff)` keyed to pulses since `lastReinforcedAt`, per-type base rate via `MEMORY_DECAY_RATES` (`pending_intention 0.04` .. `episodic 0.18`), emotional protection `d_eff = d * (1 - 0.6 * intensity)`.
- `scoreMemorySalience` swaps raw `memory.confidence` for `effectiveConfidence`; `RECENCY_WEIGHT` drops `0.5` -> `0.2` now that decay carries most of that signal.
- `shouldEvictMemory` drops a memory once it is below `EVICTION_CONFIDENCE_THRESHOLD 0.05` and past `EVICTION_MIN_AGE_PULSES 20`, exempting `unresolved`/`pending_intention`; a `MAX_MEMORIES = 500` per-agent backstop runs in `pulse-scheduler.ts`'s new `applyMemoryEviction`, called from `persistAndSnapshot` so every write-path (LLM failure, resolver failure, full commit) stays capped.
- `applyMemoryReinforcement` bumps `lastReinforcedAt` to `now` for every memory `selectRelevantMemories` surfaces and emits a `memory_reinforced` operator event (new type in `operator.types.ts`).
- `buildPerceptionPacket`/`runEngineStep` thread `now` and `pulseIntervalMs` through so decay tracks the simulated clock instead of a memory-store-relative proxy.
- Ten new/rewritten unit tests in `memory-salience.test.ts` (decay-over-pulses, emotional protection, eviction gate incl. Zeigarnik exemption, reinforcement) plus a five-case `PulseScheduler` integration suite (`pulse-scheduler-memory-decay.test.ts`, split out for the 25-tests-per-file cap): reinforcement event + timestamp bump, non-surfaced memory left alone, stale eviction, unresolved exemption, `MAX_MEMORIES` backstop.

## Why

Memory salience previously treated a memory's authored confidence as constant forever and had no eviction path, so `agentState.memories` grows unbounded (flagged as a deliberate deferral in ADR-0013) and old, no-longer-relevant memories never lost ground to genuinely current ones. #131's research resolved the model (power-law retention, per-type rate, emotional protection, Zeigarnik-exempt eviction); this closes #126 Q5 by wiring it in.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (1443 tests, 127 files)
- `node scripts/audit-tests.mjs --check` PASS, `node scripts/docs-lint.mjs --check` PASS

Closes #137

### DIFF
--- a/packages/server/src/simulation/pulse-scheduler.ts
+++ b/packages/server/src/simulation/pulse-scheduler.ts
@@ -317,7 +317,7 @@ export class PulseScheduler {
       if (engineEvents.length > 0) {
         const committed = await this.appendAndProject(engineEvents, channels, membership);
         eventsCommitted += committed.length;
-        this.applyMemoryProjection(committed, stepResult.updatedAgentState);
+        this.applyMemoryProjection(committed, stepResult.updatedAgentState, now);
       }
 
       // Reinforcement (testing effect): every memory this pulse's selection
@@ -390,7 +390,7 @@ export class PulseScheduler {
         if (resolved.committedEvents.length > 0) {
           const committed = await this.appendAndProject(resolved.committedEvents, channels, membership);
           eventsCommitted += committed.length;
-          this.applyMemoryProjection(committed, stepResult.updatedAgentState);
+          this.applyMemoryProjection(committed, stepResult.updatedAgentState, now);
         }
 
         for (const opEv of [...resolved.operatorEvents, ...runtimeOutput.operatorEvents]) {
@@ -579,8 +579,14 @@ export class PulseScheduler {
    * Mutates `updatedAgentState` in place — same-pulse `persistAndSnapshot` is
    * the single write point, and a failed append never reaches here because
    * only returned (committed) events are passed in.
+   *
+   * Both timestamps are stamped from the simulated clock (`now`), not from
+   * `event.createdAt`. Committed events carry wall-clock time; decay,
+   * reinforcement and eviction all measure age in pulses off `simTime`, so a
+   * wall-clock stamp would put every memory ~1.7e12 ms in the future of the
+   * clock reading it and nothing would ever age.
    */
-  private applyMemoryProjection(committed: CommittedEvent[], agentState: AgentState): void {
+  private applyMemoryProjection(committed: CommittedEvent[], agentState: AgentState, now: number): void {
     for (const event of committed) {
       if (event.type !== "memory_written") continue;
       agentState.memories.push({
@@ -595,8 +601,8 @@ export class PulseScheduler {
         confidence: event.payload["confidence"] as number,
         intensity: event.payload["intensity"] as number,
         unresolved: event.payload["unresolved"] as boolean,
-        createdAt: event.createdAt,
-        lastReinforcedAt: event.createdAt,
+        createdAt: now,
+        lastReinforcedAt: now,
       });
     }
   }
--- a/packages/server/src/simulation/pulse-scheduler.ts
+++ b/packages/server/src/simulation/pulse-scheduler.ts
@@ -15,7 +15,15 @@ import type {
   Memory,
 } from "@perfectman/shared";
 import { createId, createSeededRng, STAGNATION_WINDOW_PULSES, OWN_HISTORY_LIMIT } from "@perfectman/shared";
-import { runEngineStep, computeStagnationMetrics, detectAttractorStates, filterVisibleEventsForAgent } from "@perfectman/engine";
+import {
+  runEngineStep,
+  computeStagnationMetrics,
+  detectAttractorStates,
+  filterVisibleEventsForAgent,
+  effectiveConfidence,
+  shouldEvictMemory,
+  MAX_MEMORIES,
+} from "@perfectman/engine";
 import type { IEventRepository, IAgentStateRepository } from "../persistence/repositories.js";
 import type { ChannelRegistry } from "./channel-registry.js";
 import type { RateLimitGate } from "./rate-limit-gate.js";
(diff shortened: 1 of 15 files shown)


packages/server/src/simulation/pulse-scheduler.ts
```diff
@@ -523,6 +541,55 @@ export class PulseScheduler {
...
+   */
+  private async applyMemoryReinforcement(
+    surfacedMemories: readonly Memory[],
+    agentState: AgentState,
+    now: number,
+  ): Promise<void> {
+    for (const surfaced of surfacedMemories) {
+      const stored = agentState.memories.find((m) => m.id === surfaced.id);
+      if (!stored) continue;
+      stored.lastReinforcedAt = now;
```
caiotheodoro, line 625: bug: `now` here is `this.simTime` (starts at 0, `pulse-scheduler.ts:123/180`), but every memory reaching an agent's store is stamped with wall-clock time — `applyMemoryProjection` copies `event.createdAt`, and `in-memory-stores.ts:32` sets that from `Date.now()`. So `now - lastReinforcedAt` is about -1.75e12 for a freshly projected memory: `effectiveConfidence` clamps `agePulses` to 0, retention stays 1.0, and nothing ever decays or evicts on the live path.

After this line it inverts: reinforcement writes simTime into `lastReinforcedAt`, so a recalled memory is the only one that ever starts decaying. Recall causes decay instead of preventing it. It also breaks `action-intent-prompt-builder.ts:231`, which sorts staler-first on `lastReinforcedAt` and would now rank every reinforced memory below every untouched one.

Pick one clock and stamp memories with it. Simplest is to have the projection write simTime instead of `event.createdAt`, and add a scheduler test that runs real pulses off `executePulse` rather than fabricating both timestamps in the same domain — that's why the current suite passes.

caiotheodoro (commented):
One blocker inline: decay is keyed to a different clock than the timestamps it reads, so the feature is inert on the live path and inverted after recall. Also needs a rebase — the stack that just landed on main rewrote `pulse-scheduler.ts` and `memory-salience.ts` (#162 changed salience relevance), so this now conflicts in 5 files.

PR #159, opened 2026-09-05
fix(simulation): split repetition-guard blocks out of no_op_recorded

caiotheodoro:
## What

Turns a repetition-guard block into its own committed fact instead of a `no_op_recorded`:

- `repetition_blocked` event type; `IntentResolver.buildNoOpEvent` emits it when the motive carries `REPETITION_GUARD_MARKER`, `no_op_recorded` stays for chosen silence.
- `llm_retry_recovered` operator event from `ActionIntentStep` when the retry fixed `near_duplicate`/`unresolved_target`; surfaced as `ScenarioRunArtifact.recoveredFallbacks` and `BenchReport.recoveredFallbacks`.
- `PulseScheduler` stamps `lastActionAt` for a fallback no_op too (`runtimeOutput.fallbackApplied`): without relief `cold_start_bootstrap` saturates and the agent forces `needsLLM` every following pulse — two live captures, one guard block and one `No JSON object found` parse failure, both monopolized the room.
- Probes map the type to `blocked_repeat`, never `silence`; `render/html.ts` labels it `repetição bloqueada`; `sweep-repetition`/`sweep-temperature` count both shapes so older recordings still measure.
- Five tests: resolver emits `repetition_blocked`, genuine no_op unchanged, fallback no_op stamps `lastActionAt` (guard and parse-failure), bench folds recovered fallbacks into the report.

## Why

A guard block narrated as chosen silence, a degenerate run scoring as behaviorally diverse, and an agent that never got relief after a blocked turn all came from the same conflation.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+5)
- `hoc_fatia_que_nao_existe` on `deepseek/deepseek-v4-flash`: the two parse-failure no_ops in the capture now stamp `lastActionAt` (they were previously invisible as `0 fallbacks`). The stamp alone did not end that run's monopoly — `needsLLM` has other owners; tracked separately.

### DIFF
--- a/packages/eval/src/cli/bench.ts
+++ b/packages/eval/src/cli/bench.ts
@@ -57,6 +57,9 @@ export type BenchReport = {
   promptTemplateVersions: string[];
   signalsByKind: Record<string, import("../run/signal-checker.js").SignalsByKindEntry>;
   calibration: ReturnType<typeof calibrateJudge>;
+  /** Retries that fixed a guard violation on the first attempt — not free,
+   *  but not a terminal `llm_failure` either (see `llm_retry_recovered`). */
+  recoveredFallbacks: number;
   perScenario: Array<{
     id: string;
     name: string;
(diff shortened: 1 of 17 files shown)


caiotheodoro:
Merged into main as 2eb91f8 (whole stack fast-forwarded; GitHub can't retarget a fully-merged branch, so closing here).

PR #162, opened 2026-09-05
fix(engine): count self-subject memories as relevant in salience

caiotheodoro:
## What

Makes an agent's memory about itself relevant to itself:

- `scoreMemorySalience`/`selectRelevantMemories` take `selfId`; a memory whose `subjectAgentIds` include the agent earns the relevance bonus alongside `involvedPeople`; `buildPerceptionPacket` passes `agent.agentId`.
- Three tests: self-subject memory surfaces with nobody involved, self memory ranks alongside peer memories, existing callers updated to the new signature.

## Why

`involvedPeople` excludes the agent's own id by design ("who else is in this exchange"), so the "my own secret" memory shape — the hidden-objective mechanic's bread and butter — could never rank, however central it was. Overlaps `#157`'s `scoreMemorySalience` signature change; rebases after it merges.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (+3)
- Direct call with the `hoc_fatia_que_nao_existe` seeds and `involvedPeople = {iris, theo}`: both Bruno's and Marcela's secret memories surface, where before neither did.

### DIFF
--- a/packages/engine/src/perception/__tests__/memory-salience.test.ts
+++ b/packages/engine/src/perception/__tests__/memory-salience.test.ts
@@ -21,13 +21,18 @@ function memory(overrides: Partial<Memory> = {}): Memory {
   };
 }
 
+// A placeholder self id that never matches any fixture's subjectAgentIds,
+// so existing tests keep testing what they always tested unless a test
+// explicitly cares about self-relevance.
+const NOBODY = "nobody-self-id";
+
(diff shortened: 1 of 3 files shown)


caiotheodoro:
Merged into main as c7833e6 (stack fast-forwarded; GitHub can't retarget a fully-merged branch).

PR #163, opened 2026-09-05
feat(simulation): commit each intent's private motive as its own event

caiotheodoro:
## What

Turns each intent's private motive into its own committed event, joined to the act it explains:

- `IntentResolver.resolve` appends one `private_motive_summary` per resolved intent — committed, fallback-committed, blocked, delayed and `no_op` alike — `operator_only`, ordered after the act, `sourceIntentId: intent.id`, payload `PrivateMotiveSummaryPayload { summary, intentType, emotionDrivers, motivationDrivers, engineAuthored }`. Public act payloads are unchanged.
- `isEngineAuthoredMotive` / `ENGINE_MOTIVE_PREFIXES` (`packages/server/src/agent/engine-motive.ts`) become the single owner of the "is this a feeling or a parse error" convention; the narrator imports it instead of its own copy, and the payload stamps the verdict at emission.
- **Spectator gate**: `SpectatorProjection` emits `motive_reveal` only under `omniscientSpectatorMode` and the type leaves `isAlwaysSpectatorVisible` — the pre-existing case would have posted every motive to a plain spectator (and Discord) channel the moment anything emitted the type.
- `ActionIntentOperatorData.intentId` so stream consumers can join the same way.
- Two readers that counted bookkeeping as content: the scheduler's 40-event `recentEventsWindow` now excludes `operator_only` rows before slicing (one motive per intent would otherwise halve the room's memory), and `deriveDivergenceFromLog` measures over witnessable events only (it was growing with the number of agents, tipping the goal layer's meaning-made gate to `continue`).
- ADR-0014 with D-46..D-49. Ten tests: one motive per resolved intent joined by `sourceIntentId`, ordering after the act, `engineAuthored` for a fallback, motive on a blocked act, spectator null vs `motive_reveal`, engine filter drops the type, predicate on every prefix.

## Why

`motive_authenticity`, `memory_continuity` and `hidden_payoff` are scored on the private half of an act, and on a real `hoc_fatia_que_nao_existe` run that half was absent from the log for every message. The event type, the agent-side block and the spectator case already existed; nothing emitted it.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (1491, +10)
- `html-receiver-parity` e2e and the goal-layer determinism/recipe arcs pass unchanged in outcome after the two reader fixes; per-scenario committed counts rise by exactly one per LLM-resolved intent.

### DIFF
--- a/docs/adr/0014-private-motive-committed-event.md
+++ b/docs/adr/0014-private-motive-committed-event.md
@@ -1,0 +1,50 @@
+# ADR-0014: Private Motive As A Committed Event
+
+**Status**: Accepted (LOCKED, 2026-09-04) — rationale and rejected alternatives inline in this ADR (decisions **D-46..D-49**).
+
+**Numbering note**: written against the merged-tree ADRs 0001..0013; in-tree decision high-water mark **D-45** (ADR-0013). The D-46..D-49 range is this ADR's; if another unmerged slice claims the range first, renumber here and reconcile the inline D-refs.
+
+## Context
+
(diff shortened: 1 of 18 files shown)


caiotheodoro:
Merged into main as be9dd86 (stack fast-forwarded; GitHub can't retarget a fully-merged branch).

PR #165, opened 2026-09-05
feat(eval): thesis signals, act-share probe, truthful fallback counts

caiotheodoro:
## What

Encodes the hidden-objective thesis as deterministic signals and probes, and makes the fallback counts truthful:

- `ExpectedSignal` gains `forbidden_phrase_absent` (public messages by an agent, folded for case/diacritics, against the seed's `forbiddenPublicPhrases`), `private_channel_used`, `memory_referenced` (seed `secretKeywords` or memory-summary words found in the agent's joined motives, memory writes or private-channel text) and `chosen_silence_present` (LLM-resolved no-ops with a character motive); a `liveOnly` flag skips a signal in mock mode, `SignalOutcome.skipped` keeps it out of every pass rate, and `aggregateSignalsByKind` ignores it — the weekly mock gate stays at 100% while live runs are held to the thesis.
- `AgentSeedSpec.forbiddenPublicPhrases` / `.secretKeywords`; `hoc_fatia_que_nao_existe` gets both for all four agents plus the thesis signals; two re-skinned collisions on the same structure, `hoc_heranca_do_sitio` and `hoc_banda_no_festival` (three agents on one `scarceResourceId`, one scout, seeded secrets, two private channels), with golden labels and a `hoc` bench slice.
- **Probes**: `act-share-max` (largest per-agent share of content turns, band `[0, 0.5]` — the live monopoly sat at 33/64); `noop-meaningfulness` no longer counts an engine-authored motive; the adapter drops `private_motive_summary` rows from the stream and joins each motive onto its act with `engineAuthored`.
- **Runner**: `countFallbacks` counts committed and operator `llm_failure` (the parse-failure path only ever emitted the operator one, so a run with two parse fallbacks reported `0 fallbacks`); `fallbackNoOps` and `operatorEventCounts` on the artifact; `RunnerOpts.seed` threads into `localLLMConfig`; `BenchReport.signalsSkipped`.
- Fifteen tests: public vs private hit and diacritic folding, explicit phrase list, private-channel counting per agent, chosen silence excludes engine no-ops, memory keywords via motive/keywords/threshold, liveOnly skip and aggregate exclusion, act-share and its band, meaningfulness excludes engine motives, adapter join, fallback counting, and a shared contract that every constrained hoc agent declares forbidden phrases and every collision scene carries the live-only signals.

## Why

No signal could say "the secret was never said in public", "a private channel was actually used", "silence was chosen with a motive" or "the seeded memory came up" — the four things `hoc_fatia_que_nao_existe` exists to test — and a run with two parse-failure no-ops summarised itself as `0 fallbacks`.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (1517, +15)
- `bench --slice hoc --mode mock --judge rule`: 3 scenarios × variants, signal pass 100% with the live-only signals reported as skipped; `check-bench-gate.mjs` PASS on that report.


caiotheodoro:
Merged into main as 66b11a1 (stack fast-forwarded; GitHub can't retarget a fully-merged branch).

PR #166, opened 2026-09-05
feat(eval): honest calibration status, bench warnings, seeds/run-id flags

caiotheodoro:
## What

Makes the bench report honest about what it did not measure and gives it the flags an experiment needs:

- `CalibrationReport` gains `nGolden`, `nScenes` (actually matched), `nAxisPairs` and `status: "pass" | "fail" | "no_data"` with `passed: null` on no data — a kappa of 0 over zero pairs printed `FAIL` before; `calibrationVerdict` prints `NO DATA` in bench, calibrate and the evidence report.
- `BenchReport.warnings`: the narration, narration-judge and golden-narration `catch {}` blocks now record `{ scenarioId, stage, message }` and log it; a shrinking narrative sample can no longer pass unnoticed.
- `judgeAxisTargets` is the union over every selected rubric (stricter min wins) and `judgeAxisMeansByRubric` splits means by rubric id — a mixed roleplay + hidden-objective selection used to show only the first scenario's bar.
- `--seeds 42,43,44` runs each scenario once per seed (threaded to `RunnerOpts.seed`), records `perScenario[].seed` and `judgeAxisStats` (`mean`, `sd`, `min`, `max`, `n`); `--pulse-limit N` reaches `RunnerOpts.pulseLimit` for the first time; `--run-id <id>` writes `docs/eval/evidence/<id>/{bench-report.json, run-meta.json, scenarios/*.json}` and refuses an existing directory without `--force`. `run-meta.json` carries git sha, generator and judge routes, seeds, pulse cap, prompt/template versions and the *names* of the `PERFECTMAN_*` variables in force.
- Narration is scored under a jury too (primary judge config) instead of being dropped; `docs/eval/README.md` now states the temperature-0 default and the experiment flags.
- Six tests: seed expansion with per-axis spread, narration failure recorded as a warning without failing the scenario, target union across rubrics, evidence directory with env-var names only plus the `--force` refusal, and the two vacuous calibration cases now asserting `no_data`; the disagreement case matches two scenes so it measures a real `fail`.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (1521, +4)
- `bench --slice hoc --mode mock --judge rule --seeds 1,2 --run-id smoke` writes the three artifacts and prints `NO DATA` for calibration (no golden overlap in a two-run smoke) where it printed `FAIL` before.

### DIFF
--- a/docs/eval/README.md
+++ b/docs/eval/README.md
@@ -38,12 +38,18 @@ PERFECTMAN_LLM_MODEL=qwen3:8b \
 pnpm --filter @perfectman/eval bench --mode local --judge llm --out out/bench-local.json
 
 # heuristic LLM-as-judge with per-turn narrative-cohesion scoring.
-# The judge defaults to HIGH temperature (PERFECTMAN_JUDGE_TEMPERATURE) —
-# varied, creative reads expose cohesion/voice failures a strict low-temp
-# judge misses. Set it to 0 for deterministic calibration runs.
+# bench and calibrate both default the judge to temperature 0 — a judge
+# sampling creatively cannot gate a change, the same transcript scores
+# differently run to run. PERFECTMAN_JUDGE_TEMPERATURE overrides it.
 PERFECTMAN_LLM_PROVIDER=deepseek \
-PERFECTMAN_JUDGE_TEMPERATURE=1.0 \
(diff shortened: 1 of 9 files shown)


caiotheodoro:
Merged into main as fa5e08a (stack fast-forwarded; GitHub can't retarget a fully-merged branch).

PR #170, opened 2026-09-05
feat(engine): pressure discharge — acting spends the urge

caiotheodoro:
## What

Gives an urge a memory of having been acted on, so a hot persona stops feeling the same `high` pressure on every pulse:

- `AgentState.pressureDischargedAt?: Partial<Record<PressureType, number>>` — pulse index of the last committed outward act that expressed each urge (optional; absent means never discharged, so the forty fixtures that build the literal are untouched and old rows read back unchanged).
- `dischargedIntensity` in `computePressures`: `effective = raw ≥ 0.85 ? raw : max(0, raw − PRESSURE_DISCHARGE · max(0, (PRESSURE_REFRACTORY_PULSES − pulsesSince + 1) / PRESSURE_REFRACTORY_PULSES))`, applied per `ACTION_PRESSURE_MAP` entry and to the private-channel drive before the intensity label; an urge that discharges below `PRESSURE_THRESHOLD` is removed. Overwhelming urges are exempt, so ADR-0015's "overwhelming can never be inhibited" holds. `computePressures` takes `pulseIndex` (optional; `runEngineStep` passes it).
- `PulseScheduler` stamps every salient pressure at the `lastActionAt` site after a committed outward act — not only the top one, since the goulart calibration feels provoke, dominate and show-off together.
- `PRESSURE_DISCHARGE = 0.40`, `PRESSURE_REFRACTORY_PULSES = 6` (`shared/src/constants/pressure-refractory.ts`), provisional, owner #129. Worked goulart cadence: `0.73 → 0.33` the pulse after (silent unless addressed), `0.40` at +2, `0.73` again at +7.
- Persistence: nullable JSON column `pressure_discharged_at` with a guarded `ALTER TABLE` for databases created before it (only a duplicate-column error is swallowed); `AgentStateSchema` accepts the optional map.
- ADR-0016 (D-55..D-58). Seven tests: demoted next pulse and recovered after the window, overwhelming never discharged, below-threshold removal, private-channel drive discharged and the map ignored without a pulse index, scheduler stamps every salient pressure after a committed act and never for a genuine no_op, the 40-pulse two-agent run never commits messages on more than 2 consecutive pulses, schema round-trip.

## Why

`computePressures` recomputes every urge from emotional baselines each pulse; for the goulart calibration `impulsiveProvocation ≈ 0.49` is a `high urge_to_provoke` on every pulse of every run with nothing said, and no inhibition that persona has outranks it. ADR-0015 removed the ownership defects around `needsLLM`; without a memory of acting a hot persona still alternates act / cooldown / act.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (1556, +7)
- Golden mock bench and the full-registry mock bench pass `check-bench-gate.mjs` at 100% signals on this branch (after the ADR-0015 floor recalibration on the base PR); `act-share-max` mean on the golden slice drops relative to the base.

### DIFF
--- a/docs/adr/0016-pressure-discharge.md
+++ b/docs/adr/0016-pressure-discharge.md
@@ -1,0 +1,45 @@
+# ADR-0016: Pressure Discharge
+
+**Status**: Accepted (LOCKED, 2026-09-05) — rationale and rejected alternatives inline in this ADR (decisions **D-55..D-58**).
+
+**Numbering note**: written against ADRs 0001..0015; in-tree decision high-water mark **D-54** (ADR-0015). The D-55..D-58 range is this ADR's.
+
+## Context
+
+`computePressures` recomputes every urge from the agent's emotional state each pulse, with no memory of having acted on it. `computeActionEmotions` carries large baseline terms that do not depend on events: for the goulart calibration (arousal 0.65, stability 0.35) `impulsiveProvocation ≈ 0.49`, which times its pressure weight is a `high urge_to_provoke` on every pulse of every run, with nothing said. No inhibition available to that persona outranks `high`. ADR-0015 removed the ownership defects around `needsLLM`; without a memory of acting, a hot persona still alternates act / cooldown / act.
+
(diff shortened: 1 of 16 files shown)


caiotheodoro:
Merged into main as dc6c877 (stack fast-forwarded; GitHub can't retarget a fully-merged branch).

PR #171, opened 2026-09-05
feat(engine): own-history repetition window

caiotheodoro:
## What

Gives the repetition guard and the prompt's `no_repeat` block a real memory of what the agent said:

- `EngineSnapshot.ownHistoryWindow?` — the agent's own last `OWN_HISTORY_LIMIT` messages/replies from the whole committed log, built by `PulseScheduler` from the `allCommitted` it already loads (no repository change) and passed through `EngineSnapshotProjection`.
- `buildPerceptionPacket` takes it as an eighth argument and assembles `ownRecentUtterances` from the union of that window and whatever own messages are still inside the shared window: deduped by event id, chronological, consecutive duplicates collapsed, last `OWN_UTTERANCE_WINDOW`. Before this the source was the 40-event shared window alone — ~8–12 pulses of memory in a 4-agent room, which is how a pulse-18 re-announcement sailed past a guard that had seen the pulse-7 original.
- `renderNoRepeat` shows the newest `VERBATIM_UTTERANCES` (5) verbatim, as before, and older ones as one 80-char line each under "Earlier you already said (do not raise these again unless something changed)"; the token-cap trim still sheds utterances oldest-first and the step-level guard still checks the full list.
- `OWN_HISTORY_LIMIT = 12`, `OWN_UTTERANCE_WINDOW = 12` (was 5) in `shared/src/constants/repetition-window.ts`, owned by `sweep-repetition`. `single_topic_lock` stays declared-not-implemented; it belongs with the #119 stagnation re-tune.
- Five tests: an own message that fell out of the shared window is present via the own-history window; cap, newest-last, id dedupe and consecutive-duplicate collapse; the prompt renders five verbatim and the rest truncated; the scheduler hands the engine the own message after 50 foreign ones scrolled the shared window; the projection passes the field through and leaves it absent when not supplied.

## Why

On the real `hoc_fatia_que_nao_existe` run the same agent re-posted near-verbatim announcements at p7, p10, p12, p18 and p28. The Jaccard guard caught the close ones and missed p18 and p28 because its own-utterance list came from a shared 40-event window that had already scrolled past p7.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (1561, +5)
- Golden mock bench `check-bench-gate.mjs` PASS at 100% signals; `content-repetition` probe unchanged at 0 on the golden slice.

### DIFF
--- a/packages/engine/src/perception/__tests__/build-perception-packet.test.ts
+++ b/packages/engine/src/perception/__tests__/build-perception-packet.test.ts
@@ -61,3 +61,34 @@ describe("buildPerceptionPacket eventHandles", () => {
     expect(packet.eventHandles).toEqual({});
   });
 });
+
+describe("buildPerceptionPacket ownRecentUtterances", () => {
+  const own = (id: string, content: string, pulseIndex: number) =>
+    makeEvent("message_sent", { id, actorId: "agent-me", pulseIndex, createdAt: 1000 + pulseIndex, payload: { content } });
+
+  it("keeps an own message that fell out of the shared window when the own-history window still holds it", () => {
+    const old = own("evt_old", "gente, oficial: a proposta chegou", 1);
+    const foreign = Array.from({ length: 6 }, (_, i) => makeEvent("message_sent", { id: `evt_f${i}`, actorId: "agent-x", pulseIndex: 10 + i }));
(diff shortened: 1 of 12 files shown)


caiotheodoro:
Merged into main as 1719c80 (stack fast-forwarded; GitHub can't retarget a fully-merged branch).

PR #172, opened 2026-09-05
feat(simulation): seeded per-pulse turn order

caiotheodoro:
## What

Makes the per-pulse agent order a seeded shuffle instead of config order:

- `PulseScheduler.turnOrder(seed)` — Fisher–Yates over `config.agents` with `createSeededRng((seed ^ Math.imul(pulseIndex + 1, 0x9e3779b1)) >>> 0)`, a separate rng stream so the per-agent snapshot rng values are unchanged. Later agents in a pulse already see earlier agents' commits (the window refresh inside the loop), so a fixed order gave agent 0 a permanent first-mover slot and everyone else a permanent reactor slot.
- No per-pulse cap on consecutive acts here: that invariant is pinned in the engine by ADR-0015's property tests and ADR-0016's 40-pulse run; a scheduler cap would be a second owner of the gate.
- One test: the order trace over 8 pulses with 4 agents is identical for the same seed, a permutation of the cast on every pulse, not constant across pulses, and different for a different seed.

## Validation

- `pnpm lint` PASS - `pnpm test:all` PASS (1562, +1)
- Golden mock bench `check-bench-gate.mjs` PASS at 100% signals; html-receiver parity e2e passes (recorder and receiver observe the same order).

### DIFF
--- a/packages/server/src/simulation/__tests__/pulse-scheduler.test.ts
+++ b/packages/server/src/simulation/__tests__/pulse-scheduler.test.ts
@@ -916,4 +916,49 @@ describe("PulseScheduler", () => {
     });
   });
 
+
+  describe("seeded turn order", () => {
+    const FOUR = ["agent_1", "agent_2", "agent_3", "agent_4"].map((id) => ({ id, state: makeAgentState(id), persona: AGENT.persona }));
+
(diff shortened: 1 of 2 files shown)


caiotheodoro:
Merged into main as 5a9a6bc (stack fast-forwarded; GitHub can't retarget a fully-merged branch).

